PR #25, opened 2014-05-02
BuildRequire graphviz-gnome for png support

andreas-schwab:


damianatorrpm:
Sorry for a question to such an old issue, but what does the png support in libzypp mean?
I grepped in the sources for graphviz and I don't see anything.

mlandres:
It's not for the sources, it's for the documentation. Doxygen uses it to build it's images (class hierachy etc.)

damianatorrpm:
Thanks a lot for the info!

PR #40, opened 2015-01-18
PathInfo: rename major/minor to devMajor/devMinor

pinotree:
`major()` and `minor()` conlict with the GNU libc macros with the same names, so if any system header bringing them (directly or indirectly) is included before PathInfo.h, then PathInfo.h is unusable.

Unfortunately, the only clean way to solve this conflict is renaming `PathInfo::major()` & `PathInfo::minor()` resp. to `devMajor()` & `devMinor()`.

--- 
libzypp maintainers: is such kind of API change wanted? Should I bump `LIBZYPP_COMPATMINOR` & `LIBZYPP_MINOR` in VERSION.cmake as well, or will you do that just once before the release?

### DIFF
--- a/zypp/PathInfo.cc
+++ b/zypp/PathInfo.cc
@@ -236,22 +236,22 @@ namespace zypp
 
     /******************************************************************
      **
-     **	FUNCTION NAME : PathInfo::major
+     **	FUNCTION NAME : PathInfo::devMajor
      **	FUNCTION TYPE : unsigned int
      */
-    unsigned int PathInfo::major() const
+    unsigned int PathInfo::devMajor() const
     {
-      return isBlk() || isChr() ? ::major(statbuf_C.st_rdev) : 0;
+      return isBlk() || isChr() ? major(statbuf_C.st_rdev) : 0;
     }
 
     /******************************************************************
      **
-     **	FUNCTION NAME : PathInfo::minor
+     **	FUNCTION NAME : PathInfo::devMinor
      **	FUNCTION TYPE : unsigned int
      */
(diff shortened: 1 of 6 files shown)


mlandres:
Thanks for your patches. I'll have a closer look at this...

mlandres:
Do you experience any problem here?

PathInfo.h itself includes `sys/types.h`, which brings in the macros. This way we have in fact `PathInfo::gnu_dev_major()` & `PathInfo::gnu_dev_minor()` defined. That's probably not been intended, but does not require to break the API.

pinotree:
> PathInfo.h itself includes sys/types.h, which brings in the macros. This way we have in fact PathInfo::gnu_dev_major() & PathInfo::gnu_dev_minor() defined. That's probably not been intended, but does not require to break the API.

This is true, however this happens because the Linux version of GNU libc's `sys/sysmacros.h` has (simplified):
```c
unsigned int gnu_dev_major (unsigned long long int __dev);
#define major(dev) gnu_dev_major (dev)
```
but if `sys/sysmacros.h` is not implemented using inline functions but only with macros that just replace the macro parameter with the "calculated" value, then the `PathInfo.h` header becomes an invalid source.

I might guess having libzypp on non-Linux architectures (or to other Linux libc's) is not a priority, but I'd consider this "bad playing with macros" an issue to be fixed anyway, since it "slips" to direct users of the API (which are unknowning of this).

mlandres:
No doubt that renaming is right. But if there is no urgent issue with major/minor, I like to keep them as deprecated in order to avoid breaking the API now. I adapted your patch accordingly.

PR #43, opened 2015-02-06
zypp/PublicKey.cc: Use GPG_BINARY from KeyRing.cc

thp:
On systems where `gpg` is not available, checking the public key
would fail. Use `GPG_BINARY` in both `PublicKey.cc` and `KeyRing.cc`,
and define it in a single place only (`ExternalProgram.h`).


mlandres:
Basically right, but I dislike exposing such a define in a public header. For now I'd just introduce it consistently in both .cc files and add a comment for cleanup
```
/** \todo Fix duplicate define in PublicKey/KeyRing */
#define GPG_BINARY "/usr/bin/gpg2"
```

thp:
Updated PR with your suggestion.

mlandres:
Thank you

PR #45, opened 2015-05-19
Support for *.check files in the systemCheck.d directory

thp:
Reads files from `/etc/zypp/systemCheck.d` (default) in alphabetical order and adds the listed requires and conflicts rules to the system requirements as if they would be listed in the file `/etc/zypp/systemCheck`.

We use this downstream in Nemo Mobile / Sailfish OS to be able to ship custom additions to the `systemCheck` file easily.


mlandres:
Thanks.

PR #47, opened 2015-05-27
call rpmdb2solv without executing user-supplied scripts (bnc#926853)

bmwiedemann:
There is still some code duplication with ExternalProgram.cc:240 but not too bad


bmwiedemann:
added a 3rd commit that needs testing

PR #49, opened 2015-07-30
install ZyppConfig.cmake module instead of findzypp

### LINKED ISSUE #28: Install FindZypp.cmake -> /usr/<lib>/cmake/LibZypp/LibZyppConfig.cmake
The cmake module for Zypp currently gets installed to

/usr/share/cmake/Modules/FindZypp.cmake

However, this breaks compliancy with cmake's packaging howto [1]. Package modules are not supposed to be installed into /usr/share/cmake[-]/.

The correct installation path and filename is:

/usr/lib/cmake/Zypp/ZyppConfig.cmake

Also the ZyppCommon.cmake file should be stored in that very same folder.

Greets,
Mike

Updated 2018-05-28: Corrected the installation path.

xantares:
fixes #28
ZyppConfig includes ZyppCommon automatically
and also add a targets file


mlandres:
Thanks for cleaning it up. 
I'm sorry, but it will take some time until I can consider merging this (and the -bindings). Other zypp based packages are affected by the change as well. Can't afford testing or breaking our build queue atm.

Conan-Kudo:
@mlandres Could you consider reviving this for libzypp 17.0?

PR #52, opened 2015-11-17
Downloader: allow to skip appdata files

seife:
For a majority of the users, downloading huge appdata files is a total
waste of time and bandwidth, because they are only used by appdata-aware
desktop apps (GNOME). Allow users to opt-out of this by setting
ZYPP_SKIP_APPDATA=1 environment variable.

In a perfect world, we could set this in zypp.conf, but this will do for now.

### DIFF
--- a/zypp/misc/CheckAccessDeleted.cc
+++ b/zypp/misc/CheckAccessDeleted.cc
@@ -89,13 +89,10 @@ namespace zypp
         do { ++ch; } while ( *ch != '\0' );	// skip to next field
       }
 
-      if ( pinfo.command.size() == 15 )
-      {
-        // the command name might be truncated, so we check against /proc/<pid>/exe
-        Pathname command( filesystem::readlink( Pathname("/proc")/pinfo.pid/"exe" ) );
-        if ( ! command.empty() )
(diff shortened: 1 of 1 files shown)


mlandres:
* not only susetags, also rpmmd/yum repos download appdata (type="appdata"/"app-icons")
* libzypp/zypper also is appdata aware (-t application), so we can not skip appdata.xml.gz itself. Only icons could be omitted.
* additional appdata*failed/html stuff downloaded with susetags repos is due to a bug when building the media. Those files do not belong there. Hopefully fixed in https://build.opensuse.org/request/show/344935

If we allow skipping the icons, there should be an AUTO mode, which omits download IFF  no appdata consuming plugin is installed:
```
#include "zypp/ZConfig.h"
#include "zypp/PathInfo.h"

if ( filesystem::is_empty_dir( ZConfig::instance().pluginsPath()/"appdata" ) != -1 )
  // we have NO appdata plugins installed
```
The environment variable should be mentioned in `doc/autoinclude/EnvironmentVariables.doc`. The description  should mention that after installing  an appdata plugin in AUTO mode, a  forced refresh of all repos might be required to get all icons.

PR #62, opened 2016-07-12
use more user friendly command names for "zypper ps"

seife:
fixes https://bugzilla.opensuse.org/show_bug.cgi?id=980541

### DIFF
--- a/zypp/misc/CheckAccessDeleted.cc
+++ b/zypp/misc/CheckAccessDeleted.cc
@@ -89,13 +89,10 @@ namespace zypp
         do { ++ch; } while ( *ch != '\0' );	// skip to next field
       }
 
-      if ( pinfo.command.size() == 15 )
-      {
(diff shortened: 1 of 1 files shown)


mlandres:
You're right. The 'c' format option passed to 'lsof -F...' is now obsolete; and the 'c' case in the switch above you change as well. If you'd remove both, I'll merge the change.

seife:
Hi Michael,
I was not 100% sure about that, because now pinfo.command might be undefined in case readlink (for whatever reason, I don't know if this can happen in reality) returns empty. I also have no idea if an empty/undefined pinfo.command is harmless or not :-)

mlandres:
You are right Stefan, 
some system processes and kernel threads may indeed have dangling links. It does not do harm, but in this case the 'c' value is probably better than en empty entry. I'll amend my commit...

Issue #63, opened 2016-12-23
Compilation error reported by GCC7


marxin:
Error message:

```
/home/abuild/rpmbuild/BUILD/libzypp-16.3.1/zypp/sat/SolvableType.h: In member function 'bool zypp::sat::SolvableType<Derived>::isKind() const':
/home/abuild/rpmbuild/BUILD/libzypp-16.3.1/zypp/sat/SolvableType.h:66:65: error: expected primary-expression before '>' token
       bool  isKind() const    { return satSolvable().isKind<TRes>(); }
                                                                 ^
/home/abuild/rpmbuild/BUILD/libzypp-16.3.1/zypp/sat/SolvableType.h:66:67: error: expected primary-expression before ')' token
       bool  isKind() const    { return satSolvable().isKind<TRes>(); }
```

mlandres:
Fixed via #64. Thanks.

PR #64, opened 2016-12-23
Fix compilation error reported by GCC7 (#63).

marxin:


marxin:
Will you please add the patch to zypp:Head/libzypp package? Or should I do a submit request?

PR #68, opened 2017-05-29
Provide $snapshot_version repo variable.

boombatower:
The idea behind this is to support the, soon to be deployed, historical Tumbleweed snapshots in a clean manor at the libzypp level. The variable will be updated in the zypp.conf to point to the desired snapshot. I have developed a cli tool to manually replace the URLs without this change, but I will adapt to utilize this if accepted.

The end result is something like, `libypp.conf`:

```ini
snapshotVersion = 20170516
```

and a `tw-oss.repo`:

```ini
[tw-oss]
name=tw-oss-$snapshot_version
baseurl=http://download.opensuse.org/tumbleweed/$snapshot_version/repo/oss/
```

### DIFF
--- a/doc/autoinclude/RepoVariables.doc
+++ b/doc/autoinclude/RepoVariables.doc
@@ -30,4 +30,7 @@ The variable expander also supports shell like definition of default and alterna
 \li \c $releasever_major -
 	\c $releasever_major will be set to the leading portion up to (but not including) the 1st dot; \c $releasever_minor to the trailing portion after the 1st dot. If there's no dot in \c $releasever, \c $releasever_major is the same as \c $releasever and \c $releasever_minor is empty.
 
+\li \c $snapshot_version -
+	Tumbleweed snapshot version to target.
+
 */
--- a/tests/repo/RepoVariables_test.cc
+++ b/tests/repo/RepoVariables_test.cc
@@ -202,6 +202,8 @@ BOOST_AUTO_TEST_CASE(replace_text)
   BOOST_CHECK_EQUAL( replacer1("${releasever_major}"),	"13" );
   BOOST_CHECK_EQUAL( replacer1("${releasever_minor}"),	"2" );
 
+  BOOST_CHECK_EQUAL( replacer1("${snapshot_version}"),	"current" );
+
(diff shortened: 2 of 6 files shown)


mlandres:
Great idea. Would you mind generalizing it a bit?

Rather than defining `snapshotVersion` in `zypp.conf`, add a new config directory path there. You can follow `ZConfig::vendorPath()` as an example (zypp.conf and zypp/ZConfig.{h,cc}). Add `Pathname varsPath() const;` with `configPath()/"vars.d"` as the default value and 'varsdir' as the name of the zypp.conf variable to define an alternate value.

The idea is to allow a file `ZConfig::varsPath()/"snapshot_version"` to define the value of `$snapshot_version` (everything up to the first NL or EOF). So you can easily define and use arbitrary custom variables.

If you like to provide the patch for the zypp.conf, I'll adapt value retrieval and caching in `RepoVariables.cc`.

boombatower:
> Great idea. Would you mind generalizing it a bit?

I had that thought, but was not sure if would be a larger deviation and undesirable. Glad we're on the same page as I like the idea of arbitrary variables more.

> If you like to provide the patch for the zypp.conf, I'll adapt value retrieval and caching in RepoVariables.cc.

Not entirely sure what you mean. Are you indicating that the current PR could be accepted and you will rework to make it generic? If so I think that makes sense since you're likely a lot more familiar with the code. Otherwise, let me know.

mlandres:
> Are you indicating that the current PR could be accepted and you will rework to make it generic? 

Yes, I'll rework it. It's a good sketch as you touched all the relevant places.

boombatower:
Nice, I'll try out your commit in the next week or so. A side-note, but related, is there a way for an external tool to get the value of `varsdir` without having to parse `zypp.conf`? Like `git conf $varname`.

boombatower:
I updated my [`tumbleweed-cli`](https://github.com/boombatower/tumbleweed-cli) to utilize the `vars.d` functionality and it works nicely as expected. Thanks for implementing the feature!

Now just need to finally have the snapshots themselves deployed in production.

mlandres:
I added a 'zypper config' to thew wish list as there is by now no tool to inspect the `zypp.conf' settings.

PR #69, opened 2017-05-29
Set dupAllowVendorChange to 'false' by default

sysrich:
Based on discussions triggered after my recent Tumbleweed talk ( https://youtu.be/RbP9lNvmWKk?t=32m16s ) and the realisation of the following:

1. Tumbleweed obviously already only recommend dup with --no-allow-vendor-change
2. SLE's recommendation for 'dup' usage is only with --no-allow-vendor-change (https://www.susecon.com/doc/2015/sessions/TUT16735.pdf slide 44)
3. A dup from Leap versions with --no-allow-vendor-change would allow a smoother transition between minor versions while still using OBS repos and simplify the guidance on https://en.opensuse.org/SDB:System_upgrade

This in my mind makes a compelling argument for the default for dupAllowVendorChange to by false.

### DIFF
--- a/zypp.conf
+++ b/zypp.conf
@@ -368,19 +368,18 @@
 
 ##
 ## EXPERTS ONLY: TUNE DISTRIBUTION UPGRADE (DUP)
-## Set whether to allow changing the packages vendor upon DUP. If you
-## are following a continuous distribution like Tumbleweed or Factory
-## where you use 'zypper dup --no-allow-vendor-change' quite frequently,
-## you may indeed benefit from disabling the VendorChange. Packages from
-## OBS repos will then be kept rather than being overwritten by Tumbleweeds
-## version.
+## Set whether to allow changing the packages vendor upon DUP. In most cases you
+## want to continue using the packages from the same source/vendor as currently
+## installed. If you would prefer Packages from OBS/additional repos overwriting
+## packages from your currently in-use repositories you may consider changing
+## this option.
 ##
 ## CHANGING THE DEFAULT IS NOT RECOMMENDED.
 ##
 ## Valid values:  boolean
-## Default value: true
+## Default value: false
 ##
-# solver.dupAllowVendorChange = true
+# solver.dupAllowVendorChange = false
 
 ##
 ## EXPERTS ONLY: Cleanup when deleting packages. Whether the solver should
--- a/zypp/ZConfig.cc
+++ b/zypp/ZConfig.cc
@@ -331,7 +331,7 @@ namespace zypp
 	, solver_dupAllowDowngrade	( true )
 	, solver_dupAllowNameChange	( true )
 	, solver_dupAllowArchChange	( true )
-	, solver_dupAllowVendorChange	( true )
+	, solver_dupAllowVendorChange	( false )
         , solver_cleandepsOnRemove	( false )
         , solver_upgradeTestcasesToKeep	( 2 )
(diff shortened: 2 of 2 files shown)


mlandres:
No. Mainly because the current branch is used on already released products back to 42.2/SLE12sp2. We can't simply change the builtin defaults without explicit consent of all stakeholders.

But maybe it's time to introduce some distribution/product specific config file, overwriting the builtin defaults and being overwritten by /etc/zypp/zypp.conf. This could then be maintained and shipped e.g. with the distributions -release package.

sysrich:
But SLE 12 SP2 already use --no-allow-vendor-change when using zypper dup (the cited pdf is back from SLE 12 SP1 & 2). So I think an argument can be made that all stakeholders are already aligned to this.

That said, a distirbution/product specific config file sounds like a very good thing, not just for situations like this but other ones like SLE/openSUSE JeOS where a different setting for solver.onlyRequires is currently hacked in a messy way

mlandres:
> But SLE 12 SP2 already use --no-allow-vendor-change when using zypper dup...

But this is just what `we` recommend for a migration. We do not know what 'dup' jobs customer scripts, SuMa., PackageKit , Yast, Salt, Ansible, Kiwi, etc. are using and what would break if we'd change the default behavior.

okurz:
How about accepting this PR for this repo and apply as-is for openSUSE Factory and any newer derived product branched from this. The SR for updates to any older released product could then be accompanied with a specific configuration keeping the old default intact.

mlandres:
This branch also feeds SLE12-sp2 and 3. It would prevent us from releasing fixes until  a distirbution/product specific config file is implemented and shipped for these products.

If it's that urgent, let e.g. the openSUSE-release package inject `solver.dupAllowVendorChange = false` into the zypp.conf. The file is `%config(noreplace)`, so libzypp won't change it once it is installed.

okurz:
> It would prevent us from releasing fixes until a distirbution/product specific config file is implemented and shipped for these products.

https://build.suse.de/package/show/home:okurz:branches:SUSE:SLE-12-SP2:Update/libzypp should solve this. As soon as you would update the version for the next SLE submission you can merge in this package including the reverse patch to preserve the existing SLE12 SP2/3 behaviour.

mlandres:
Zypp is not maintained that way. Authority is the git brach; building, testing and submission to obs is automated. We don't accept/support obs patches.

But even if: The sample zypp.conf just lists the available options (lines are `#`-commented). The systems actually use the built-in default from ZConfig.cc. This is the value that must not change.

Furthermore zypp.conf is %config(noreplace), so any changes in the file we ship would not affect systems with already installed and edited config files. This is no way how you'd reach all TW systems.

ATM you must ask the user to edit the zypp.conf (or forcefully do it).

okurz:
What has been proposed as well is to have a package installed read-only zypp.conf in /usr/share which is branded per distribution and make a file in /etc/ optional so that a sysadmin can still decide to have a different configuration on demand.

mlandres:
Yes, a _distribution/product specific config file_, as suggested at the beginning.

mlandres:
The issue is fixed for TW with libzypp-16.3.0 (by 9f55d8b179ad41131ecf0be54b4088a88f185b42).

okurz:
awesome work, thank you very much.

mlandres:
@DimStar77 is to blame for it ;)

PR #72, opened 2017-07-04
rename: handle EXDEV error on OverlayFS

vrothberg:
The `rename(2)` systemcall doesn't succeed across devices, which is likely
to happen when using the overlay storage driver in Docker when trying to
rename between the upper and the lower layer. In such case, errno is
set to *EXDEV*.

Extend `PathInfo::rename` to handle the *EXDEV* errno case and fallback to
using `mv(1)`, which is explicitly mentioned in the kernel docs to deal
correctly with OverlayFS issues.

**Fixes:** https://github.com/openSUSE/docker-containers/issues/57
Signed-off-by: Valentin Rothberg <vrothberg@suse.com>

### DIFF
--- a/zypp/PathInfo.cc
+++ b/zypp/PathInfo.cc
@@ -660,6 +660,34 @@ namespace zypp
     }
 
     ///////////////////////////////////////////////////////////////////
+    namespace
+    {
+      int safe_rename( const Pathname & oldpath, const Pathname & newpath )
+      {
+        int ret = ::rename( oldpath.asString().c_str(), newpath.asString().c_str() );
+
+        // rename(2) can fail on OverlayFS. Fallback to using mv(1), which is
+        // explicitly mentioned in the kernel docs to deal correctly with OverlayFS.
+        if ( ret == -1 && errno == EXDEV ) {
+          const char *const argv[] = {
+            "/usr/bin/mv",
+            oldpath.asString().c_str(),
+            newpath.asString().c_str(),
+            NULL
+          };
+          ExternalProgram prog( argv, ExternalProgram::Stderr_To_Stdout );
+          for ( string output( prog.receiveLine() ); output.length(); output = prog.receiveLine() ) {
+            MIL << "  " << output;
+          }
+          ret = prog.close();
+        }
+
+        return ret;
+      }
+    } // namespace
+    ///////////////////////////////////////////////////////////////////
+
+    ///////////////////////////////////////////////////////////////////
     //
     //	METHOD NAME : rename
     //	METHOD TYPE : int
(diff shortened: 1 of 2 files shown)


zypp/RepoManager.cc
```diff
@@ -1238,6 +1238,9 @@ namespace zypp
   void RepoManager::Impl::buildCache( const RepoInfo & info, CacheBuildPolicy policy, const ProgressData::ReceiverFnc & progressrcv )
   {
     assert_alias(info);
+    cleanCache(info);
+    refreshMetadata(info, RefreshForced);
+
```
mlandres: Why?

zypp/PathInfo.cc
```diff
@@ -661,13 +661,41 @@ namespace zypp
 
     ///////////////////////////////////////////////////////////////////
     //
+    //	METHOD NAME : safe_rename
+    //	METHOD TYPE : int
+    //
```
mlandres: `safe_rename` is considered to be an internal helper, so it should be defined inside an anonymous namespace and not be mentioned in the header file. (The 'oldfashioned ' METHOD comment block above (662-666) can be removed. We no longer do this for new code.) 
```
///////////////////////////////////////////////////////////////////
namespace
{
```

zypp/PathInfo.cc
```diff
@@ -661,13 +661,41 @@ namespace zypp
...
+        ExternalProgram prog( argv, ExternalProgram::Stderr_To_Stdout );
+        for ( string output( prog.receiveLine() ); output.length(); output = prog.receiveLine() ) {
+          MIL << "  " << output;
+        }
+        ret = prog.close();
+      }
+
+      return ret;
+    }
+
```
mlandres:
```
} // namespace
///////////////////////////////////////////////////////////////////
```

zypp/PathInfo.h
```diff
@@ -604,6 +604,14 @@ namespace zypp
...
     int rename( const Pathname & oldpath, const Pathname & newpath );
 
+    /**
+     * Like '::rename', but falls back to using mv(1) in case of EXDEV, which
+     * can occur when using OverlayFS.
+     *
+     * @return 0 on success, errno on failure
+     **/
+
+    int safe_rename( const Pathname & oldpath, const Pathname & newpath );
```
mlandres: `safe_rename` is just an internal helper and should not be exposed in the header.  The comment of `rename` could be updated to mention the fallback to `mv`.

zypp/RepoManager.cc
vrothberg: Stupid mistake. Leftover from debugging that I failed to remove from this commit. I'll fix this along with the other comments and update the PR soon.

vrothberg:
I updated the PR, thanks for the review :+1:

PR #73, opened 2017-07-13
Fix c++-11 compiler warning

ya1gaurav:
By default destructors are noexcept in c++-11 so gcc compiler will give warning.
Fix applied.

### DIFF
--- a/zypp/base/ReferenceCounted.cc
+++ b/zypp/base/ReferenceCounted.cc
@@ -30,7 +30,7 @@ namespace zypp
     : _counter( 0 )
     {}
 
-    ReferenceCounted::~ReferenceCounted()
+    ReferenceCounted::~ReferenceCounted() noexcept(false)
     {
       if ( _counter )
         {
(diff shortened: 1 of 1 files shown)


mlandres:
Your compiler does not require this to match the declaration in the header file? ;)

Anyway, the exception was meant to be a safety net as we started the development.  It's actually not needed and not usefull. I'd rather remove the exception than introducing `noexcept(false)`. If you like to change your PR accordingly, I'll take it.

ya1gaurav:
So, I remove throw statement?

mlandres:
yes

mlandres:
@ya1gaurav Please be so kind to join those 2 commits into one. Thanks.

(E.g. `git rebase -i` and `squash` the commits using `Remove throw...` as log message.  Needs `push -f` when pushing back into your branch)

ya1gaurav:
done in https://github.com/openSUSE/libzypp/pull/75

mlandres:
Thanks.

(you know that you could have fixed this in branch ya1gaurav:patch-1 without need to open a new PR)

PR #92, opened 2018-02-27
WIP Clang format

bzeller:


bzeller:
This is obsolete, even if we'd go for auto formatting the whole codebase we can do a new PR

PR #98, opened 2018-03-04
PoolQuery::addDependency add overload to define an individual Match::…

mlandres:
…Mode (bsc#1043166)

JFYI: The enhancement here will be needed to fix zypper search (bsc#1043166), that's why we refer to the bug number here in libzypp too. The commit policy for SLES requires a bug reference in the changelog to accept a submission. If this PR happens to be the only change in libzypp, we would otherwise not be able to release it in order to provide a fixed zypper.

### DIFF
--- a/zypp/PoolQuery.cc
+++ b/zypp/PoolQuery.cc
@@ -229,7 +229,7 @@ namespace zypp
 
       /** Dumb serialization.
        * \code
-       *   AttrMatchData ATTRIBUTE SEARCHSTRING [C|X] SERIALIZED_PREDICATE
+       *   AttrMatchData ATTRIBUTE SEARCHSTRING [C|SEARCHMODE|X] SERIALIZED_PREDICATE
        * \endcode
       */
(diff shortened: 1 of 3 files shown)


zypp/PoolQuery.cc
```diff
@@ -316,6 +313,56 @@ namespace zypp
...
+    private:
+      /** Serialize \ref strMatcher \ref Match::Mode */
+      static std::string serializeMode( Match::Mode mode_r )
+      {
+	// Legacy code used "[C|X]" to differ just between OTHER (need to (C)ompile) and
+	// using the default search mode. As we now allow to specify a SEARCHMODE we
+	// need to serialize it:
+	switch ( mode_r )
+	{
+#define OUTS(M,S) case Match::M: return #S; break
```
bzeller: I don't think the #define here adds a lot in terms of readability. IMHO defines should be avoided as much as possible, they can be hard to read and especially they are hard to debug.

PR #103, opened 2018-03-15
Add MediaSetAccess::provideFileFromUrl

bzeller:


zypp/PublicKey.h
```diff
@@ -331,8 +334,6 @@ namespace zypp
     friend class KeyRing;
     /** KeyRing ctor: No need to parse file if KeyRing already had valid KeyData. */
     PublicKey( const filesystem::TmpFile & sharedFile_r, const PublicKeyData & keyData_r );
-    /** KeyRing ctor: Legacy callback APIs take PublicKey, but just need the PublicKeyData No need to export to file. */
-    explicit PublicKey( const PublicKeyData & keyData_r );
```
mlandres: Wanted to cherry-pick this commit, but PublicKey.h change is not related (if necessary at all).

bzeller: That was a leftover from my tests with loading all keys manually with KeyManager and trying to register them manually. Cleaned the branch up to only contain related changes

PR #107, opened 2018-03-23
Make sure the file returned by provideFileFromUrl is not deleted befo…

bzeller:
…re use

Fixes a bug in the newly introduced provideFileFromUrl in MediaSetAccess and adds a small testcase to prevent from breaking the behaviour in the future.

### DIFF
--- a/zypp/MediaSetAccess.cc
+++ b/zypp/MediaSetAccess.cc
@@ -16,6 +16,7 @@
 #include "zypp/ZYppCallbacks.h"
 #include "zypp/MediaSetAccess.h"
 #include "zypp/PathInfo.h"
+#include "zypp/TmpPath.h"
 //#include "zypp/source/MediaSetAccessReportReceivers.h"
 
 using namespace std;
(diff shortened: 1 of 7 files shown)


zypp/MediaSetAccess.cc
```diff
@@ -170,13 +171,23 @@ IMPL_PTR_TYPE(MediaSetAccess);
...
+    filesystem::TmpFile tmpFile;
+    tmpFile.autoCleanup(false); //cleaned up by ManagedFile
+
+    Pathname file = access.provideFile(path, 1, options);
+
+    //prevent the file from being deleted when MediaSetAccess gets out of scope
+    if ( filesystem::hardlinkCopy(file, tmpFile.path()) != 0 )
+      ZYPP_THROW(Exception("Can't copy file from " + file.asString() + " to " +  tmpFile.path().asString() ));
+
+    return ManagedFile ( tmpFile.path(), filesystem::unlink );
```
mlandres: IMO not exception safe. `tmpFile.autoCleanup(false)` must not be called before the `ManagedFile` actually manages it.

I'd almost consider providing a `static ManagedFile TmpFile::asManagedFile();` in `TmpPath.h/.cc` (as separate commit). Then there's no need for the code to fiddle with the `TmpFile` at all.

bzeller: Ouch yes, Done

mlandres:
The `TmpFile::asManagedFile` (zypp/TmpPath.*)  is a feature on it's own. It should go into s separate commit.  Then it's good to be merged.

PR #113, opened 2018-06-08
Ma fix fetcher checks

mlandres:
Rewrote Fetcher to move only validated files to the final destination.  A file already found at the final destination, but not validating, will be unlinked.

This behavior is desired to fix [bsc#1091624](https://bugzilla.suse.com/show_bug.cgi?id=1091624). Package signature check will be moved to the Fetcher, so we don't get unchecked Packages into the cache.

### DIFF
--- a/zypp/Fetcher.cc
+++ b/zypp/Fetcher.cc
@@ -27,8 +27,6 @@
 #include "zypp/parser/susetags/ContentFileReader.h"
 #include "zypp/parser/susetags/RepoIndex.h"
 
-using namespace std;
-
 #undef ZYPP_BASE_LOGGER_LOGGROUP
 #define ZYPP_BASE_LOGGER_LOGGROUP "zypp:fetcher"
 
@@ -108,7 +106,7 @@ namespace zypp
     OnMediaLocation location;
     Pathname deltafile;
     //CompositeFileChecker checkers;
-    list<FileChecker> checkers;
+    std::list<FileChecker> checkers;
     Flags flags;
   };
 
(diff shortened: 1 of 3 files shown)


zypp/Fetcher.cc
```diff
@@ -535,59 +503,58 @@ namespace zypp
...
-
-    // start look in cache
-    got_from_cache = provideFromCache(resource, dest_dir);
+    const OnMediaLocation & resource( jobp_r->location );
 
-    if ( ! got_from_cache )
+    try
     {
-      MIL << "Not found in cache, downloading" << endl;
+      shared_ptr<MediaSetAccess::ReleaseFileGuard> releaseFileGuard; // will take care provided files get released
```
bzeller: scoped_ptr should be enough here ( or std::unique_ptr )

zypp/Fetcher.cc
```diff
@@ -332,77 +329,48 @@ namespace zypp
...
-    // call the checker function
     try
     {
-      MIL << "Checking job [" << localfile << "] (" << checkers.size() << " checkers )" << endl;
+      MIL << "Checking job [" << localfile_r << "] (" << checkers_r.size() << " checkers )" << endl;
 
-      for ( list<FileChecker>::const_iterator it = checkers.begin();
-            it != checkers.end();
-            ++it )
+      for ( FileChecker chkfnc : checkers_r )
```
bzeller: Maybe use FileChecker & , could be the std::function carries more than just a function pointer, e.g. lambda with lots of variables in scope , or a functor object.

zypp/Fetcher.cc
mlandres: Right. The initial guard used a custom deleter, but no longer need it.

zypp/Fetcher.cc
mlandres: Changed to `const &`

PR #114, opened 2018-06-08
Rpmhead download

bzeller:
Teaches libzypp to download the RPM header of a package when the required information is not available. This makes it possible to see filelists and changelog from a not yet installed or upgradable package.

### DIFF
--- a/zypp/MediaSetAccess.cc
+++ b/zypp/MediaSetAccess.cc
@@ -171,6 +171,25 @@ IMPL_PTR_TYPE(MediaSetAccess);
     return op.result;
   }
 
+  Pathname MediaSetAccess::provideFileHead(const OnMediaLocation &resource_r, const ByteCount minBytes_r, ProvideFileOptions options_r)
+  {
+    Pathname result;
+    auto op = [ &minBytes_r, &result ] ( media::MediaAccessId media, const Pathname &file ) {
+      media::MediaManager media_mgr;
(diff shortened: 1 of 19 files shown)


Conan-Kudo:
@bzeller Umm, why not just fetch the filelist/other metadata instead?

bzeller:
@Conan-Kudo not sure we have that detailed information in libsolv. But even if we do we only save a few bytes of download. We can always optimize further if we think that it is required :)

Conan-Kudo:
@bzeller The information is available in libsolv, as we use it in libdnf through libsolv. libzypp just needs to request parsing it.

bzeller:
@Conan-Kudo also currently the code needs to have the full rpm header at least for the filelist case, we try to not break ABI here. So we decided to stay with full header download. But I will keep your infos in mind in case we really see need to improve on that.

Conan-Kudo:
@bzeller You don't already download the filelist.xml content instead?

bzeller:
@Conan-Kudo no, we download only the RPM header for the requested package. The filelist is too big to download on demand , we'd need a own cache handling for that like apt-file does for example. That is out of scope for this PR, here we just wanted to extend the information you can get when querying info for a specific package. In case we ever introduce a feature to search through all available files in enabled repositories we could switch over using the available information there instead.

bzeller:
Obsolete, if that feature would come it would be part of the zypp-ng code stream

PR #115, opened 2018-06-08
Check for not imported keys after multi key import from rpmdb

bzeller:
After importing multiple keys from rpmdb check if all of them were expected. If not show it in the logs.

### DIFF
--- a/zypp/target/rpm/RpmDb.cc
+++ b/zypp/target/rpm/RpmDb.cc
@@ -1005,10 +1005,23 @@ void RpmDb::syncTrustedKeys( SyncTrustedKeyBits mode_r )
     try
     {
       getZYpp()->keyRing()->multiKeyImport( tmpfile.path(), true /*trusted*/);
+      // bsc#1096217: Try to spot and report legacy V3 keys found in the rpm database.
+      // Modern rpm does not import those keys, but when migrating a pre SLE12 system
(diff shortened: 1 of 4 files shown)


zypp/target/rpm/RpmDb.cc
```diff
@@ -1005,10 +1008,18 @@ void RpmDb::syncTrustedKeys( SyncTrustedKeyBits mode_r )
     try
     {
       getZYpp()->keyRing()->multiKeyImport( tmpfile.path(), true /*trusted*/);
+      for ( const PublicKey &key : getZYpp()->keyRing()->trustedPublicKeys() ) {
+        expectedKeys.erase( key.gpgPubkeyEdition() );
+      }
+
+      for ( const Edition &notImportedKey : expectedKeys ) {
+        ERR << "Could not import key:" << str::Format("gpg-pubkey-%1%-%2%") % notImportedKey.version() % notImportedKey.release() << " into zypp keyring" << endl;
```
mlandres: Not sure if this is right. AFAIR (need to check it), if the rpm db contains multiple releases of the same key FFFFFFFF-11111111 and FFFFFFFF-22222222, the keyring will show just FFFFFFFF-22222222 (the latest release). 
If so, we just need to warn if a VERSION is lost (but not on a RELEASE).
But basically this is the location where we can trigger a JobReport(or a geneeric keyringCB) so zypper is able to pritnt some hint

PR #120, opened 2018-06-18
Replace 'redhat-release' with generic 'system-release'

Conan-Kudo:
fedora-release and mageia-release offer generic `system-release` Provides that can be used to determine the release version of the distribution. Derivatives of Fedora (including RHEL) also offer this generic capability in their distribution release packages.

This mirrors the change that was done in libsolv last year to support more distributions correctly in the `solv` example program.


mlandres:
Thanks

PR #121, opened 2018-06-19
Ma adjust sigchecks

mlandres:
- no error reporting, no tmpfile when only retrieving the fingerprints
- restore the old (GPG2) behavior: accept valid sig even if the key is meanwhile expired

### DIFF
--- a/zypp/KeyManager.cc
+++ b/zypp/KeyManager.cc
@@ -68,9 +68,27 @@ public:
   Impl();
   ~Impl();
 
-  std::list<std::string> verifyAndReadSignaturesFprs(const Pathname & file, const Pathname & signature, bool &verifed);
+  /** Return all fingerprints found in \a signature_r. */
+  std::list<std::string> readSignaturesFprs( const Pathname & signature_r )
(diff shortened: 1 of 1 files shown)


zypp/KeyManager.cc
```diff
@@ -68,9 +68,27 @@ class KeyManagerCtx::Impl
...
 
-  std::list<std::string> verifyAndReadSignaturesFprs(const Pathname & file, const Pathname & signature, bool &verifed);
+  /** Return all fingerprints found in \a signature_r. */
+  std::list<std::string> readSignaturesFprs( const Pathname & signature_r )
+  { return readSignaturesFprsOptVerify( signature_r ); }
+
+  /** Tries to verify the \a file_r using \a signature_r. */
+  bool verifySignaturesFprs( const Pathname & file_r, const Pathname & signature_r )
+  {
+    bool verify;
```
bzeller: I know the called function does initialize it but it makes me nervous to see a uninitialized variable like that :D. Not a must but = false here will not hurt.

mlandres: done

PR #122, opened 2018-06-22
Ma fix pkgsigcheck

mlandres:
Fixing [bsc#1091624](https://bugzilla.suse.com/show_bug.cgi?id=1091624)

The packages sig check code was moved into a FileChecker passed down to the Fetcher. This avoids the package being moved into the cache before the check has completed.

### DIFF
--- a/zypp/repo/PackageProvider.cc
+++ b/zypp/repo/PackageProvider.cc
@@ -523,17 +523,10 @@ namespace zypp
 
     ManagedFile RpmPackageProvider::doProvidePackage() const
     {
-      Url url;
-      RepoInfo info = _package->repoInfo();
-      // FIXME we only support the first url for now.
-      if ( info.baseUrlsEmpty() )
-        ZYPP_THROW(Exception("No url in repository."));
-      else
-        url = * info.baseUrlsBegin();
-
       // check whether to process patch/delta rpms
+      // FIXME we only check the first url for now.
       if ( ZConfig::instance().download_use_deltarpm()
-	&& ( url.schemeIsDownloading() || ZConfig::instance().download_use_deltarpm_always() ) )
+	&& ( _package->repoInfo().url().schemeIsDownloading() || ZConfig::instance().download_use_deltarpm_always() ) )
       {
 	std::list<DeltaRpm> deltaRpms;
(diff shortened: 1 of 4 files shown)


mlandres:
TODO:
- [ ]  Need to fix the case of using delta rpms to build the rpm localy. They must be built outside the cache, then be checked before moving them to the cache. Will add one more commit for this.

zypp/repo/PackageProvider.cc
```diff
@@ -189,22 +275,20 @@ namespace zypp
...
-	    _retry = true;
-	    break;
 	  case repo::DownloadResolvableReport::IGNORE:
 	    WAR << _package->asUserString() << ": " << "User requested to accept insecure file" << endl;
 	    break;
 	  default:
+	  case repo::DownloadResolvableReport::RETRY:
 	  case repo::DownloadResolvableReport::ABORT:
-	    ZYPP_THROW(AbortRequestException("User requested to abort"));
+	    ZYPP_THROW(RpmSigCheckException(action_r,"UFF..."));
```
bzeller: UFF..... ? ;D

mlandres:
Final commit added. TODOs are done.

zypp/repo/PackageProvider.cc
```diff
@@ -586,18 +579,28 @@ namespace zypp
...
       report()->finishDeltaApply();
 
-      return ManagedFile( destination, filesystem::unlink );
+      // Check and move it into the cache
+      // Here the rpm itself is ready. If the packages sigcheck fails, it
+      // makes no sense to return a ManagedFile() and fallback to download the
+      // full rpm. It won't be different. So let the exceptions escape...
+      rpmSigFileChecker( builddest );
+      if ( filesystem::hardlinkCopy( builddest, cachedest ) != 0 )
+	ZYPP_THROW( Exception( str::Str() << "Can't hardlink/copy " << builddest << " to " << cachedest ) );
```
bzeller: That line should be intented

bzeller: Actually it seems it is, but with a \t rather than spaces. The Github diff view does not show that properly.

PR #125, opened 2018-07-03
lsof: use '-K i' if lsof supports it (bsc#1099847)

mlandres:


mlandres:
@bzeller I don't see a build being scheduled? Or just slow...

zypp/misc/CheckAccessDeleted.cc
```diff
@@ -74,6 +75,29 @@ namespace zypp
 
       ino_t pidNS;
     };
+
+    /** bsc#1099847: Check for lsof version < 4.90 which does not support '-K i'
+     * Just a quick check to allow code15 libzypp runnig in a code12 environment.
+     */
+    bool lsofNoOptKi()
```
bzeller: if the function is only used in the current file we can mark it static

mlandres: I'm using annon. namespaces rather than statics. The function can't be addressed  from outside the compilation unit.

bzeller: Ok learnt something new , good to merge then

PR #126, opened 2018-07-04
Prevent the system from sleeping during a commit

bzeller:
(fixes openSUSE/zypper#135)

This patch prevents the system from shutting down or hibernating when a zypper commit is running.

### DIFF
--- a/zypp/ExternalProgram.h
+++ b/zypp/ExternalProgram.h
@@ -225,10 +225,12 @@ namespace zypp {
       /** Remember execution errors like failed fork/exec. */
       std::string _execError;
 
+    protected:
+
       void start_program (const char *const *argv, const Environment & environment,
     			Stderr_Disposition stderr_disp = Normal_Stderr,
     			int stderr_fd = -1, bool default_locale = false,
-    			const char* root = NULL);
+    			const char* root = NULL, bool switch_pgid = false);
 
     };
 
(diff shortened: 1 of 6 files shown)


zypp/ShutdownLock_p.h
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Attempts to create a lock to prevent the system
+ * from going into hibernate/shutdown. The lock is automatically
+ * released when the object is destroyed.
+ */
+class ZYPP_LOCAL ShutdownLock
+{
+public:
+   ShutdownLock( const std::string reason );
```
mlandres: IMO either `const std::string &`  or plain `std::string` if you want to store/move it.

bzeller: Good catch, seems I missed that

zypp/ExternalProgram.h
```diff
@@ -299,6 +301,22 @@ namespace zypp {
       std::string _buffer;
   };
 
+  /** ExternalProgram extended to change the progress group ID after forking.
+   * \see \ref ExternalProgram
+   */
+  class ExternalProgramWithSeperatePgid : public ExternalProgram
```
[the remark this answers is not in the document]
bzeller: Done

Conan-Kudo:
Why is this necessary? RPM already will inhibit when transactions are running...

mlandres:
> RPM already will inhibit when transactions are running...
This does not cover the download phase, especially not if download and install are alternating.

PR #128, opened 2018-07-27
Add filesize check for downloads with known size (bsc#408814)

bzeller:
This patch adds a check of downloaded data size in cases where we know the expected filesize beforehand.


Conan-Kudo:
@mlandres @bzeller Can we have this backported to the SLE 12-SP2 branch? I need it for closing out [rh#1615232](https://bugzilla.redhat.com/show_bug.cgi?id=1615232) on Fedora 27, which is using libzypp 16.17.20 currently.

mlandres:
@Conan-Kudo Backports will be tracked in the corresponding [bsc#408814](https://bugzilla.suse.com/show_bug.cgi?id=408814). SLE 12-SP2/3 is on the list.

PR #131, opened 2018-08-16
Fix build with Boost 1.68.0

AdamMajer:
As of Boost 1.68.0, the System library headers were updated
which breaks the hack for inlining some functions in
older Boost versions. Use proper define starting with this
Boost version before including Boost.System headers.

This only seems to affect one test.

### DIFF
--- a/tests/lib/WebServer.cc
+++ b/tests/lib/WebServer.cc
@@ -1,9 +1,13 @@
-
 #include <sstream>
 #include <string>
+#include "boost/version.hpp"
+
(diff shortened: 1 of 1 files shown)


AdamMajer:
without patch:
https://build.opensuse.org/package/show/home:adamm:boost_test/libzypp

patched:
https://build.opensuse.org/package/live_build_log/home:adamm:branches:zypp:Head/libzypp/boost_test/x86_64

mlandres:
Thanks for the patches

PR #135, opened 2018-10-12
Make sure gettext macros are not redefined

bzeller:
I get a lot of warnings in the new command PRs about the Gettext macros being redefined. This patch will make sure those warnings will no longer pop up

### DIFF
--- a/zypp/base/Gettext.h
+++ b/zypp/base/Gettext.h
@@ -14,6 +14,17 @@
 #ifndef ZYPP_BASE_GETTEXT_H
 #define ZYPP_BASE_GETTEXT_H
 
+/** Undef translation macros if they already exist */
+#ifdef _
+#undef _
+#endif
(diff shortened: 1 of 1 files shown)


zypp/base/Gettext.h
```diff
@@ -14,6 +14,17 @@
 #ifndef ZYPP_BASE_GETTEXT_H
 #define ZYPP_BASE_GETTEXT_H
 
+/** Undef translation macros if they already exist */
```
mlandres: IMO we should differ between the `#ifdef ZYPP_DLL` and `#else` case below.

- `#define N_(MSG)` should be moved into both cases

then

- In `#ifdef ZYPP_DLL`  we can unconditionally `#undef` each macro before `#define`ing it
- In the `#else` case we may define them if they are not already defined (by the application).

PR #147, opened 2018-11-26
Do not fail service-refresh on a empty repoindex.xml (bnc#1116840)

bzeller:


zypp/repo/ServiceRepos.cc
```diff
@@ -40,10 +40,17 @@ namespace zypp
...
+	try {
+	  parser::RepoindexFileReader reader(path, callback);
+	  service.setProbedTtl( reader.ttl() );	// hack! Modifying the const Service to set parsed TTL
+	  mediamanager.release( mid );
+	  mediamanager.close( mid );
+	} catch ( const Exception &e ) {
+	  //Reader throws a bare exception, we need to translate it into something our calling
+	  //code expects and handles (bnc#1116840)
+	  ZYPP_CAUGHT ( e );
+	  ZYPP_THROW( ServicePluginInformalException ( e.msg() ) );
```
mlandres: We should use [Exception::remember](https://github.com/openSUSE/libzypp/blob/master/zypp/base/Exception.h#L216) to store the original Exception as history.

PR #155, opened 2019-01-21
Try to the missing rpmDB compat symlink in case the rpm package got d…

mlandres:
…eleted ([bsc#1122471](https://bugzilla.suse.com/show_bug.cgi?id=1122471)). Changing the rpmdb path in libzypp would not help, as tools refer to /var/lib/rpm as well.

### DIFF
--- a/zypp/target/rpm/RpmDb.cc
+++ b/zypp/target/rpm/RpmDb.cc
@@ -59,6 +59,9 @@ using namespace zypp::filesystem;
 
 #define WORKAROUNDRPMPWDBUG
 
+#undef ZYPP_BASE_LOGGER_LOGGROUP
+#define ZYPP_BASE_LOGGER_LOGGROUP "librpmDb"
+
 namespace zypp
 {
   namespace zypp_readonly_hack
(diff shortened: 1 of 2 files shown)


zypp/target/rpm/RpmDb.cc
```diff
@@ -331,6 +334,14 @@ void RpmDb::initDatabase( Pathname root_r, Pathname dbPath_r, bool doRebuild_r )
     ZYPP_THROW(RpmInvalidRootException(root_r, dbPath_r));
   }
 
+  if ( dbPath_r == "/var/lib/rpm"
+    && ! PathInfo( root_r/"/var/lib/rpm" ).isExist()
+    && PathInfo( root_r/"/usr/lib/sysimage/rpm" ).isDir() )
+  {
+    WAR << "Rpm package was deleted? Injecting missing rpmdb compat symlink." << endl;
+    filesystem::symlink( "../../usr/lib/sysimage/rpm", root_r/"/var/lib/rpm" );
```
bzeller: Why do you use a relative path here for the symlink and not the one you checked for in line 339?

zypp/target/rpm/RpmDb.cc
```diff
@@ -331,6 +334,14 @@ void RpmDb::initDatabase( Pathname root_r, Pathname dbPath_r, bool doRebuild_r )
     ZYPP_THROW(RpmInvalidRootException(root_r, dbPath_r));
   }
 
+  if ( dbPath_r == "/var/lib/rpm"
+    && ! PathInfo( root_r/"/var/lib/rpm" ).isExist()
```
bzeller: Could use root_r / dbPath_r

zypp/target/rpm/RpmDb.cc
bzeller: Also we could use root_r / dbPath_r instead of the hard coded string

zypp/target/rpm/RpmDb.cc
mlandres: Yes. I just wanted to strengthen the fact that the code is not intended to be used with arbitrary dbPath_r.

zypp/target/rpm/RpmDb.cc
mlandres: Symlinks must be relative otherwise they do not work in rooted environments. The ocation must be valid no matter if or where the system is rooted to.

bzeller: Makes sense, thanks for clarifying

PR #160, opened 2019-02-07
Don't blame user on TargetAbortedException (fixes bnc#978193)

bzeller:


.gitignore
```diff
@@ -9,4 +9,4 @@ build
 .externalToolBuilders
 .settings
 *flymake.h.gch
-*.user
+CMakeLists.txt.user*
```
mlandres: .gitignore change is unrelated should go into a separate commit.

bzeller: Done

PR #164, opened 2019-03-14
Provide more information about problems

bzeller:
Libsolv offers more information about problems than we currently show
to the user. This patch enables libzypp to get the full set of
information. Also it removes some duplicate code from libzypp that is
already present in libsolv.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,117 @@
+---
+Language: Cpp
+AccessModifierOffset: -2
+AlignAfterOpenBracket: DontAlign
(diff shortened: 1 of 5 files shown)


.clang-format
```diff
@@ -0,0 +1,117 @@
+---
```
mlandres: If you want to add `.clang-format`, please in separate commit.

bzeller: It already is a seperate commit ?

mlandres: I stand corrected.

PR #166, opened 2019-04-01
Fix the openSUSE package when using CMake >= 3.14

### LINKED ISSUE #365: Commit plugins: Try to move COMMITBEGIN after download
Download errors or `--download-ony` should not trigger the commit plugins. Snappers plugin e.g. creates a btrfs snapshot, which is actually not needed if no action is performed.

krop:
Starting with CMake 3.14, EXCLUDE_FROM_ALL now spreads from
directories to targets. 'make -C someSubdir' when 'someSubdir'
uses the 'EXCLUDE_FROM_ALL' keyword does nothing.

Fixes #365


krop:
Build log with these changes and CMake 3.14: https://build.opensuse.org/package/show/home:cgiboudeaux:branches:openSUSE:Factory:Rings:1-MinimalX/libzypp

PR #176, opened 2019-07-10
KeyManagerCtx: rewrite API to take the keyring when constructing the context

mlandres:
Preparing the context and setting up the keyring must succeed, otherwise the calling methods are lost. `createForOpenPGP` was changed to prepare context and keyring in one go and to throw a `KeyRingException` on error.

Also dropped the outer `shared_ptr<KeyManagerCtx>` and let `createForOpenPGP` return the `KeyManagerCtx` directly.

The `KeyManager.cc` diff looks more complicated than the change actually is: 
- `Impl` stores the gpgme context, so it is now also responsible for releasing it.
- `createForOpenPGP` was changed to return a `KeyManagerCtx`, throw on error and immediately sets the homedir/keyring.
- `setHomedir()` was dropped.
- `listKeys()` temporarily sets `GPGME_KEYLIST_MODE_SIGS`.
- `readKeyFromFile()`  works around bsc#1140670 by using a temp. keyring if necessary (a version using `gpgme_op_keylist_from_data_start` once the bug is fixed is sketched)

### DIFF
--- a/zypp/KeyManager.cc
+++ b/zypp/KeyManager.cc
@@ -7,11 +7,12 @@
 |                                                                      |
 \---------------------------------------------------------------------*/
 #include "zypp/KeyManager.h"
-#include "zypp/PublicKey.h"
+#include "zypp/KeyRing.h"
 #include "zypp/PathInfo.h"
 #include "zypp/base/Logger.h"
 #include "zypp/TmpPath.h"
 #include "zypp/base/String.h"
+#include "zypp/AutoDispose.h"
 
 #include <boost/thread/once.hpp>
 #include <boost/interprocess/smart_ptr/scoped_ptr.hpp>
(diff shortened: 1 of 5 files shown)


zypp/KeyManager.cc
```diff
@@ -216,114 +219,153 @@ std::list<std::string> KeyManagerCtx::Impl::readSignaturesFprsOptVerify( const P
...
+  // gpgme_data_t and gpgme_op_keylist_from_data_start. Despite GPGME_KEYLIST_MODE_SIGS
+  // the signatures are missing, but we need them to create proper PublicKeyData objects.
+  // While this is not resolved, we read into a temp. keyring. Impl::_volatile helps
+  // to detect whether we can clear and import into the current context or need to
+  // create a temp. one.
+  std::list<PublicKeyData> ret;
+
+  if ( _pimpl->_volatile ) {
+    // in a volatile context we can simple clear the keyring...
+    filesystem::clean_dir( homedir() );
```
bzeller: I wonder if its a good idea to do this. It kind of is abusing undefined behaviour of what gpgme does when the homedir() is cleared. Upgrading gpgme could change the behaviour and break that part. 
Did you try to use a statically created KeyManagerCtx for all the volatile cases and purge it instead after the op is done?

mlandres: > All cryptographic operations in GPGME are performed within a context, which contains the internal state of the operation as well as configuration parameters. By using several contexts you can run several cryptographic operations in parallel, with different configuration. 

GPGME is and must be prepared for concurrent access to the homedir from the same process as well as from others. Also any GPGME operation creates a missing keyring on demand in case it is needed (not when a context is created). The behavior is IMO not that undefined (and not much different from using the keyring dir the 1st time).

Deleting all the imported keys again is also possible, but far slower.
(and as soon as bsc#1140670 is fixed, we don't need this hack at all)

PR #178, opened 2019-07-18
Drop all remaining libhal support. It's not used since there is udev.

mlandres:
Unused since SLE12, so we can remove it in master.


mlandres:
Added 2 more commits eliminating zypps own `thread::callOnce` and `thread::Mutex` implementations. It's all in std:: meanwhile :)

PR #183, opened 2019-08-29
Misc fixes

dcermak:
This is a relatively big PR of miscellaneous fixes: most of these fix warnings that are raised by gcc and clang when run with `-Wall -Wextra`.

ca4491b is pretty huge as it comments out all the unused function parameters, I can move it to a separate PR to make this easier to review.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,4 +1,4 @@
----
+
 Language: Cpp
 AccessModifierOffset: -2
 AlignAfterOpenBracket: DontAlign
@@ -74,7 +74,6 @@ IndentWidth: 2
 IndentWrappedFunctionNames: false
 JavaScriptQuotes: Leave
 JavaScriptWrapImports: true
-KeepLineBreaksForNonEmptyLines: false
 KeepEmptyLinesAtTheStartOfBlocks: true
 MacroBlockBegin: ''
 MacroBlockEnd: ''
--- a/tests/parser/HistoryLogReader_test.cc
+++ b/tests/parser/HistoryLogReader_test.cc
@@ -4,17 +4,6 @@
 
 using namespace zypp;
 
-namespace
-{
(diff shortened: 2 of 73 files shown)


bzeller:
Wow thanks for the big effort you put into that, but seems that some tests broke because of the changes. Once those are fixed it looks good to me. @mlandres should have a second look because its a really big PR.

mlandres:
No time for a proper review before my vacation, but actually I dislike [Mass comment out unused function parameters](https://github.com/openSUSE/libzypp/pull/183/commits/ca4491b62d4c44d3ff7121eb2fd25269aa1f0d4d). Sorry for all the work, but I don't see the benefit (just patch conflicts in backports).

For [Remove unused functions from tests](https://github.com/openSUSE/libzypp/pull/183/commits/b9520ccb94a02440a6ad8fd97358cd5230e963e8) an `#if 0` would be ok, as I use those methods when debugging and testing, but not necessarily in production.

dcermak:
> No time for a proper review before my vacation, but actually I dislike [Mass comment out unused function parameters](https://github.com/openSUSE/libzypp/pull/183/commits/ca4491b62d4c44d3ff7121eb2fd25269aa1f0d4d). Sorry for all the work, but I don't see the benefit (just patch conflicts in backports).

It makes it more explicit that certain parameters are not used by a function at all. That can serve as an indicator when one should refactor code (e.g. if the `MediaCD` class should go away, then there would be no more child classes of `MediaHandler` that actualy use the `ejectDev` parameter in `releaseFrom()`).

> 
> For [Remove unused functions from tests](https://github.com/openSUSE/libzypp/pull/183/commits/b9520ccb94a02440a6ad8fd97358cd5230e963e8) an `#if 0` would be ok, as I use those methods when debugging and testing, but not necessarily in production.

I can mark them with `__attribute__((unused))` instead.

PR #184, opened 2019-10-08
New container detection algorithm for zypper ps

bzeller:
New algorithm that detects if a given PID is running inside a container.
The previous one just took the PID namespace into account, however that
could be misleading since containers can share the PID namespace with
the Host.

Fixes bsc#1146947


bzeller:
> It would be nit-picky to complain about a few missing ` ` after `//` ;)
Still fixed it now that I saw them ;)

PR #187, opened 2019-10-17
Expose pool_whatmatchesdep from zypp::sat::Pool

bzeller:
Extend zypp::sat::Pool by wrapping the pool_whatmatchesdep
functionality from libsolv. This is required to implement a reverse dependency search
in zypper (fixes openSUSE/zypper#214)

### DIFF
--- a/zypp/sat/Pool.cc
+++ b/zypp/sat/Pool.cc
@@ -119,6 +119,27 @@ namespace zypp
     Pool::SolvableIterator Pool::solvablesEnd() const
     { return SolvableIterator(); }
 
+    sat::Queue Pool::whatMatchesDep( const SolvAttr &attr, const Capability &cap ) const
+    {
+      sat::Queue q;
+      pool_whatmatchesdep( get(), attr.id(), cap.id(), q, 0);
+      return q;
+    }
+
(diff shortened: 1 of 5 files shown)


zypp/sat/Pool.cc
```diff
@@ -119,6 +119,16 @@ namespace zypp
     Pool::SolvableIterator Pool::solvablesEnd() const
     { return SolvableIterator(); }
 
+    sat::Queue Pool::whatMatchesDep(const SolvAttr &attr, const Capability &cap) const
+    {
+      std::list<Solvable> matchedSolvables;
```
mlandres: Appears to be unused.

mlandres:
We could add `pool_whatmatchessolvable` and  `pool_whatcontainsdep` the same way. Sooner or later we find a usecase for them.

zypp/sat/Pool.cc
bzeller: Leftover of how the code previously worked, removed

bzeller:
> We could add `pool_whatmatchessolvable` and `pool_whatcontainsdep` the same way. Sooner or later we find a usecase for them.

Seperate commit for each?

mlandres:
No, can be all in one.

PR #189, opened 2019-11-13
Introduce PurgeKernels class (bsc#1155198)

bzeller:
Adds libzypp API to mark all obsolete kernels according to the existing
purge-kernel script rules

### DIFF
--- a/tests/repo/RepoLicense_test.cc
+++ b/tests/repo/RepoLicense_test.cc
@@ -11,7 +11,15 @@ using namespace zypp;
 using std::cout;
 using std::endl;
 
-TestSetup test( Arch_x86_64 );
+static TestSetup test( TestSetup::initLater );
+struct TestInit {
+  TestInit() {
+    test = TestSetup( Arch_x86_64 );
+  }
+  ~TestInit() { test.reset(); }
+};
+BOOST_GLOBAL_FIXTURE( TestInit );
+
 const Pathname DATADIR( TESTS_SRC_DIR "/repo/RepoLicense" );
 
 ///////////////////////////////////////////////////////////////////
--- a/zypp/PurgeKernels.cc
+++ b/zypp/PurgeKernels.cc
@@ -1,0 +1,422 @@
+/*---------------------------------------------------------------------\
+|                          ____ _   __ __ ___                          |
+|                         |__  / \ / / . \ . \                         |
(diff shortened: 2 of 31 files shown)


tests/repo/RepoLicense_test.cc
```diff
@@ -11,15 +11,24 @@ using namespace zypp;
 using std::cout;
 using std::endl;
 
-TestSetup test( Arch_x86_64 );
+
+static std::unique_ptr<TestSetup> testSetup;
```
mlandres: Basically right, but I'd prefer a solution that requires less syntactical changes in the existing code (no var rename, no var to ptr).
The `std::unique_ptr` can be used as [`_pimpl` for TestSetup](https://github.com/openSUSE/libzypp/blob/ma-draft/tests/lib/TestSetup.h). A testcase intro could then look like [this](https://github.com/openSUSE/libzypp/blob/ma-draft/tests/repo/RepoLicense_test.cc#L14). Usage of `test` in existing testcase would not need to be changed.

bzeller: Done

zypp/PurgeKernels.cc
```diff
@@ -0,0 +1,422 @@
...
+    const str::regex specRegex( "^(latest|oldest)([+-][0-9]+)?$", str::regex::match_extended );
+
+    const unsigned tokenGrp = 1; //index of the group matching the token
+    const unsigned modifierGrp = 2; //index of the group matching the offset modifier
+
+
+    MIL << "Parsing keep spec: " << _keepSpec << std::endl;
+
+    std::vector<std::string> words;
+    str::split( _keepSpec, std::back_inserter(words), "," );
```
mlandres: I'd relax the parser to trim the words. Actually I'd enhance str::split to accept an optional `Trim` argument:
```
    template<class TOutputIterator>
      unsigned split( const C_Str & line_r, TOutputIterator result_r, const C_Str & sepchars_r = " \t", const Trim trim_r = NO_TRIM )
    template<class TOutputIterator>
      unsigned split( const C_Str & line_r, TOutputIterator result_r, const Trim trim_r )
      { return split( line_r, result_r, " \t", trim_r ); }
```
May cherry-pick b4343234febc303c6f655e6b2905c4346cb49b11

bzeller: Done

PR #191, opened 2019-11-21
Pass correct posttrans script argument

bzeller:


mlschroe:
But is that correct? The argument is actually the package count.

bzeller:
> But is that correct? The argument is actually the package count.

Hm right, its  "1" for a install or "2 or higher (depending on the number of versions installed) " for a update. I guess we could count the installed versions of a package too. But aside from multiversion kernel packages it won't make a difference.

mlschroe:
I do not get a '2' when doing an update, as the old version is deinstalled at the end of the transaction. Who said that update will result in '2'?

bzeller:
> I do not get a '2' when doing an update, as the old version is deinstalled at the end of the transaction. Who said that update will result in '2'?

Yes you are right, i was referring to https://docs.fedoraproject.org/en-US/Fedora_Draft_Documentation/0.1/html/RPM_Guide/ch09s04s05.html  but after reading the rpm code and some testing it seems to be the nr of packages that are installed after the transaction ends. New code should do that.

zypp/target/RpmPostTransCollector.cc
```diff
@@ -52,7 +52,7 @@ namespace zypp
 	{ if ( !_scripts.empty() ) discardScripts(); }
 
 	/** Extract and remember a packages %posttrans script for later execution. */
-	bool collectScriptFromPackage( ManagedFile rpmPackage_r )
+        bool collectScriptFromPackage( ManagedFile rpmPackage_r, const std::string &scriptArg )
```
mlandres: Why do we still need `scriptArg` when collecting the script inside the transaction?

According to the pastebin I posted on IRC, installing 3 versions of a package, all 3 %posttrans get a 3 as argument. So it's the number of instances installed when executing the script, not when collecting it.

bzeller: Because when the scripts are executed all we know is the path to the script. 
But I guess we could just save the Pooltem::name together with the script?

mlandres: Yes. Can be done _in_ `collectScriptFromPackage`. It rerieves the `RpmHeader` anyway, so you can get and save the pkg name.

bzeller: Done

PR #195, opened 2019-12-09
Introduce new history type for patch state changes

bzeller:


zypp/HistoryLog.cc
```diff
@@ -334,6 +334,27 @@ namespace zypp
     }
   }
 
+  void HistoryLog::patchStateChange( const PoolItem & pi, const std::string &oldstate )
+  {
+    const Patch::constPtr p = asKind<Patch>(pi.resolvable());
+    if (!p)
+      return;
```
mlandres: This should work: `if ( ! pi.isKind<Patch>() ) return;`
The _'conversion to pointer and cast'_ style test is how we did it ~10 years ago :)

If there are more ancient test styles in the file, yoi can fix them on the fy too.

zypp/PoolItem.h
```diff
@@ -104,6 +104,9 @@ namespace zypp
 
       /** Broken (needed) but locked patches. */
       bool isUnwanted() const;
+
+      std::string statusName () const;
```
mlandres: I'd disambiguate it, so it does not get confused with the PoolItems status. And a conversion to string is usually soemthing with asString, so `patchStatusAsString` ore something like this,.

zypp/HistoryLog.cc
bzeller: Ok, but it won't make any difference in lines though because i need to cast to Patch anyway ;)

zypp/PoolItem.h
bzeller: Thought about this, but wasn't completely sure its only used for patches , will change it

PR #202, opened 2020-01-20
yum::Downloader: Cleanup implementation and support zchnk

mlandres:
Lots of cleanup and removing unneeded code (e.g. patches.xml parsing is Code10/11 legacy).

The initial commits fix OnMediaLocation and RepomdFileReader, parsing and storing the additional zchnk data from the repomd.xml.

The later commits fix the yum::Downloader to prefer downloading the zchnk files. (curently disabled 
by d7fe57b because libsolv chokes on it, at leest in my test environment).

Maybe adding another commit on top, removing the ResourceType class (passed from parser to downloader), which seems to be no longer needed.


zypp/repo/yum/Downloader.cc
```diff
@@ -30,109 +28,119 @@ namespace repo
...
-{
-  OnMediaLocation loc_with_path(loc_with_path_prefix(loc_r, repoInfo().path()));
-  MIL << id_r << " : " << loc_with_path << endl;
-  this->enqueueDigested(loc_with_path,  FileChecker(), search_deltafile(_delta_dir + "repodata", loc_r.filename()));
-  return true;
-}
-
-
-//bool repomd_Callback2( const OnMediaLocation &loc, const ResourceType &dtype, const std::string &typestr, UserData & userData_r );
+    // search old repository file file to run the delta algorithm on
```
bzeller: file file

zypp/repo/yum/Downloader.cc
```diff
@@ -30,109 +28,119 @@ namespace repo
...
+    Pathname search_deltafile( const Pathname & dir, const Pathname & file )
+    {
+      Pathname deltafile;
+      if ( ! PathInfo(dir).isDir() )
+	return deltafile;
+
+      std::string base = file.basename();
+      size_t hypoff = base.find( "-" );
+      if ( hypoff != std::string::npos )
+	base.replace( 0, hypoff + 1, "" );
```
bzeller: why is it required to replace the string before the first hyphen in the filename?

zypp/repo/yum/Downloader.cc
```diff
@@ -30,109 +28,119 @@ namespace repo
...
     {
       addWantedLocale( ZConfig::instance().textLocale() );
       for ( const Locale & it : ZConfig::instance().repoRefreshLocales() )
 	addWantedLocale( it );
     }
 
-    /** The callback invoked by the RepomdFileReader */
-    bool repomd_Callback2( const OnMediaLocation & loc_r, const ResourceType & dtype_r, const std::string & typestr_r )
+    /** The callback invoked by the RepomdFileReader.
+     * It's a pitty, but in the presence of separate "type" and "type_zck" entries,
```
bzeller: *nitpick* : pitty -> pity

zypp/repo/yum/Downloader.cc
mlandres: done

zypp/repo/yum/Downloader.cc
mlandres: done

zypp/repo/yum/Downloader.cc
mlandres: `6eff3456ea6b110cb95e825b91d6b62b6e83bd6ab074649baec7f7831fda646e-primary.xml`
Strips the checksum preceding the file stem so we can look for an old `*-primary.xml` which may contain some reusable blocks.

PR #203, opened 2020-01-20
Zchunk support

bzeller:
Implement zchunk support for InputStream, first steps to support zchunk in repositories.


mlandres:
:( We need to amend the SPEC/CMAKE file: libzstd is available on leap15 but not on sle15 

```
# Please submit bugfixes or comments via http://bugs.opensuse.org/
#

%if 0%{?is_opensuse} && (0%{?sle_version} >= 150000 || 0%{?suse_version} >= 1500)
%bcond_without zchunk
%else
%bcond_with zchunk
%endif

....

%if %{with zchunk}
BuildRequires:  libzck-devel
%endif

...

cmake ...
	%{?with_zchunk:-DENABLE_ZCHUNK_COMPRESSION=1} 
```


```
OPTION (ENABLE_ZCHUNK_COMPRESSION "Build with zchunk compression support?" OFF)

...

IF (ENABLE_ZCHUNK_COMPRESSION)
  ...
  ADD_DEFINITIONS (-DENABLE_ZCHUNK_COMPRESSION=1)
ENDIF (ENABLE_ZCHUNK_COMPRESSION)
```

(For the Downloader I can fortunately check whether libsolv supports ZCHUNK. If not we need to provide the plain files)

PR #205, opened 2020-01-30
SLE-5116 log patch status changes in history

mlandres:
[SLE-5116](https://jira.suse.com/browse/SLE-5116)
Basically ResPool stores the computed patch status whenever the pool content changes (repo add/remove). Upon commit patches were the status is changed by the commit are remembered in the history.
Adapted the history type for patch status changes a bit.

Ready for review, just the last commit (https://github.com/openSUSE/libzypp/commit/932874f4ead400f5e335822a1d1978a25b6799a3) will need to be tuned in case the commit is aborted. In this case some patches may not have completed their status change. Need to figure out which and omit them in the history.

### DIFF
--- a/zypp/ResStatus.cc
+++ b/zypp/ResStatus.cc
@@ -62,6 +62,37 @@ namespace zypp
     }
   }
 
+  namespace
+  {
+    // NOTE: Strings defined here are written to the history file
+    // and used by zypper. Do not change them!
+    static const char* strBROKEN = "needed";
+    static const char* strSATISFIED = "applied";
+    static const char* strNONRELEVANT = "not-needed";
+    static const char* strUNDETERMINED = "undetermined";
+  }
+
+  std::string ResStatus::validateValueAsString( ValidateValue val_r )
+  {
+    const char* ret = strUNDETERMINED;
+    switch ( val_r )
+    {
+      case BROKEN:	ret = strBROKEN; break;
+      case SATISFIED:	ret = strSATISFIED; break;
+      case NONRELEVANT:	ret = strNONRELEVANT; break;
+      case UNDETERMINED:break;// to avoid default:
+    }
+    return ret;
+  }
+
+  ResStatus::ValidateValue ResStatus::stringToValidateValue( const std::string & str_r )
+  {
+    ValidateValue ret = UNDETERMINED;
+    if      ( str_r == strBROKEN )      ret = BROKEN;
+    else if ( str_r == strSATISFIED )   ret = SATISFIED;
+    else if ( str_r == strNONRELEVANT ) ret = NONRELEVANT;
+    return ret;
+  }
 
   /******************************************************************
   **
--- a/zypp/solver/detail/SATResolver.cc
+++ b/zypp/solver/detail/SATResolver.cc
@@ -441,7 +441,7 @@ SATResolver::solving(const CapabilitySet & requires_caps,
 	  {
 	    if ( id < 0 )
 	      continue;
-	    sat::Solvable slv { id };
+	    sat::Solvable slv { (sat::detail::SolvableIdType)id };
 	    // get product buddies (they carry the weakremover)...
 	    static const Capability productCap { "product()" };
 	    if ( slv && slv.provides().matches( productCap ) )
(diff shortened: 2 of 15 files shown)


zypp/solver/detail/SATResolver.cc
```diff
@@ -178,6 +179,43 @@ int vendorCheck( sat::detail::CPool *pool, Solvable *solvable1, Solvable *solvab
...
+  {
+    MIL << "Establish..." << endl;
+    sat::detail::CPool * cPool { sat::Pool::instance().get() };
+    ::pool_set_custom_vendorcheck( cPool, &vendorCheck );
+
+    sat::Queue jobQueue;
+    AutoDispose<sat::detail::CSolver*> cSolver { ::solver_create( cPool ), ::solver_free };
+    sat::Pool::instance().prepare();
+    if ( ::solver_solve( cSolver, jobQueue ) != 0 )
+      INT << "How can establish fail?" << endl;
```
bzeller: After a error like that does it make sense to even continue here?

zypp/solver/detail/SATResolver.cc
```diff
@@ -90,6 +90,74 @@ namespace zypp
...
+	    if ( traits::isPseudoInstalled( pi.kind() ) ) ret.push( pi.id() );
+	  return ret;
+	}
+
+	/** Copy back new \ref WeakValue to \ref PoolItem after solving.
+	 * On the fly collect orphaned items (cached by the solver for the UI)
+	 */
+	inline void solverCopyBackWeak( sat::detail::CSolver & satSolver_r, PoolItemList & orphanedItems_r )
+	{
+	  #warning CHECK CODE and assert all items weak stati are reset (resetWeak called)
```
bzeller: Is that planned to stay? Because it will always emit a warning when the file is compiled... and will get lost in all the warnings. A comment might do just as well..

zypp/ResStatus.cc
```diff
@@ -62,6 +62,37 @@ namespace zypp
...
+      case SATISFIED:	ret = strSATISFIED; break;
+      case NONRELEVANT:	ret = strNONRELEVANT; break;
+      case UNDETERMINED:break;// to avoid default:
+    }
+    return ret;
+  }
+
+  ResStatus::ValidateValue ResStatus::stringToValidateValue( const std::string & str_r )
+  {
+    ValidateValue ret = UNDETERMINED;
```
bzeller: No need for a temporary var, you could directly return the value too...

zypp/ResStatus.cc
```diff
@@ -62,6 +62,37 @@ namespace zypp
...
+    // and used by zypper. Do not change them!
+    static const char* strBROKEN = "needed";
+    static const char* strSATISFIED = "applied";
+    static const char* strNONRELEVANT = "not-needed";
+    static const char* strUNDETERMINED = "undetermined";
+  }
+
+  std::string ResStatus::validateValueAsString( ValidateValue val_r )
+  {
+    const char* ret = strUNDETERMINED;
```
bzeller: No need for a temporary var, you could directly return the value too...

zypp/solver/detail/SATResolver.cc
mlandres: There are no jobs, so the error shouldn't happen (would be kind of 'internal solver error'). Just to be sure.

zypp/solver/detail/SATResolver.cc
mlandres: removed

zypp/ResStatus.cc
mlandres: Allow copy elision. I often use this pattern without thinking about whether the size if the returned object would fit into a register. The compilers RVO eliminates the temporary.

zypp/ResStatus.cc
mlandres: Allow copy elision.

mlandres:
> Some small comments, there is still a fixup! commit in there that needs to be rebased :)

We can't rebase it, as 8b5e9f3b6fb4fd85081b6417ab59aa7d0f171109 is already released.

mlandres:
Finished 'Log patch status changes to history (jsc#SLE-5116)' to reload the target in case of an aborted transaction.

Rewrote 'ResPool: add changedPseudoInstalled() to retrieve patches changed by..'. Reloading the target (changes the pool) triggers a new establish. So we need to be able to save the initial patch status before reoading and then compare against the status with he new target.

PR #207, opened 2020-02-05
WebServer::start: Wait up to 10 seconds and return whether the port i…

mlandres:
…s now active

### DIFF
--- a/tests/lib/WebServer.cc
+++ b/tests/lib/WebServer.cc
@@ -4,6 +4,14 @@
 #include <poll.h>
 #include <signal.h>
 
+#include <stdlib.h>
+#include <unistd.h>
+#include <string.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <netdb.h>
+
 #include "zypp/base/Logger.h"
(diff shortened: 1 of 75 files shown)


tests/lib/WebServer.cc
```diff
@@ -417,9 +459,20 @@ WebServer::WebServer(const Pathname &root, unsigned int port, bool useSSL)
...
+    TriBool ret { checkLocalPort( _pimpl->port() ) };
+    if ( ret != true )
+    {
+      unsigned i = 0;
+      do {
+	sleep( 1 );
+	ret = checkLocalPort( _pimpl->port() );
+      } while ( ret != true || ++i == 10 );
+    }
+    return ret;
```
bzeller: Can we move that part into the worker_thread() main instead of the std::this_thread::sleep_for(std::chrono::milliseconds(100)); In case the port never opens its the correct place to clean up the webserver process and stop the thread right away. 

The start() should not return a TriBool it either started or not. WebServer::isStopped() already checks if it was started successful.

mlandres: done

PR #219, opened 2020-04-02
Bug 1166610   gcc 10 libzypp build fails

mlandres:
[bsc#1166610](https://bugzilla.opensuse.org/show_bug.cgi?id=1166610)


zypp/zyppng/base/abstracteventsource.h
```diff
@@ -14,9 +14,9 @@
 #ifndef ZYPP_NG_BASE_ABSTRACTEVENTSOURCE_H_INCLUDED
 #define ZYPP_NG_BASE_ABSTRACTEVENTSOURCE_H_INCLUDED
 
-#include <zypp/zyppng/base/zyppglobal.h>
-#include <zypp/zyppng/base/Signals>
-#include <zypp/zyppng/base/Base>
+#include "zypp/zyppng/base/zyppglobal.h"
+#include "zypp/zyppng/base/Signals"
+#include "zypp/zyppng/base/Base"
```
bzeller, line 20: Why did you change the includes to "" ? As by definiton this is only different from the <> style includes in the way that it first searches the current directory for the given file and then the include path, which makes no sense to use if we give the full zypp/zyppng/.... path since we do not want to include from the local dir:
https://en.cppreference.com/w/cpp/preprocessor/include

I did use the <> throughout all off zyppng headers and cc files so far because i think its semantically wrong to use "" .

mlandres: We use everywhere ´<>´ for system includes and ´""´ for ours. ´"´ (AFAIK) does not search in the current directory, it searches in the directory where the including file was found. I.e. you preferably stay within your include tree. (even if we use zypp/...).
Similar zypper includes <zypp/...> but "Zypper.h".

bzeller: From doc: " Typical implementations first search the directory where the current file resides and, only if the file is not found, search the standard include directories as with (1)."  so if I include "zypp/a/b.h" and in b.h i include "zypp/c/x.h"  it would first look for "zypp/a/zypp/c/x.h" then continue with the std include dirs ...

mlandres: So it would be reasonable to change it the other way round. all `""` into `<>`...

bzeller: I think so yes

mlandres: done

PR #234, opened 2020-07-16
disable kernel multiversion setting (boo#1172229)

lnussel:


zypp/PurgeKernels.cc
```diff
@@ -255,14 +255,19 @@ namespace zypp {
...
+    }
 
     MIL << "Parsing keep spec: " << _keepSpec << std::endl;
 
     std::vector<std::string> words;
     str::split( _keepSpec, std::back_inserter(words), ",", str::TRIM );
     if ( words.empty() ) {
-      WAR << "Invalid keep spec: " << _keepSpec << " using default latest,running." << std::endl;
-      return;
+      WAR << "Invalid keep spec: " << _keepSpec << " using default latest,latest-1,running." << std::endl;
```
bzeller, line 267: Should we really just fall back to the default keep spec if the admin speficied a incorrect one? That might result in purging kernels and packages that should not be purged. IMO its better to error out in that case.

lnussel: maybe but that would be separate change. The previous code just skipped the parsing in error case so used the global defaults of those variables. Ie equivalent to "latest,running". That's even less than the config file default of "latest,latest-1,running". So this change is for consistency.

mlandres:
@lnussel boo#1172229 is about transactional systems. Your patch here would change everything down to SLE/LEAP 15.0. I don't think we want this.

I'll try provide a way to ship an empty zypp.conf in the future, keeping our current defaults, unless all TW and SLE15 stakeholders conclude that no-multiversion kernel shall be the new default for all those distros from now on.
Then transactional systems will be able to disable multiversion kernels by installing a file in multiversion.d (and the admin may overwrite this in zypp.conf again if he likes to).

lnussel:
> @lnussel boo#1172229 is about transactional systems. Your patch here would change everything down to SLE/LEAP 15.0. I don't think we want this.

Eh, don't you have branches? :)

> I'll try provide a way to ship an empty zypp.conf in the future, keeping our current defaults, unless all TW and SLE15 stakeholders conclude that no-multiversion kernel shall be the new default for all those distros from now on.
> Then transactional systems will be able to disable multiversion kernels by installing a file in multiversion.d (and the admin may overwrite this in zypp.conf again if he likes to).

Yes, that's the idea. So far I was thinking of moving the config to the purge-kernel-service but the maintainer had concerns that the setting and the service might not necessarily related. So instead we could create a subpackge of libzypp or zypper for that.

Conan-Kudo:
This change should be done at image composition time, no? And for transactional-server role in YaST, can't we include an action there to set this properly?

PR #239, opened 2020-08-17
Improve Italian traslation of the "breaking dependencies" message.

dfaggioli:
In Italian, "to break" can indeed be translated with "interrompere". But
that depends on the context, and this one here is really not the case.

Adopt a sentence which, albeit not a literal translation, sounds a lot
more natural and easier to understand.

FTR, this was reported (by me) as a bug here:
 https://bugzilla.opensuse.org/show_bug.cgi?id=1173529

### DIFF
--- a/po/it.po
+++ b/po/it.po
@@ -4439,7 +4439,7 @@ msgstr "Impossibile creare sat-pool."
 #: zypp/solver/detail/ProblemSolutionIgnore.cc:42
 #, c-format, boost-format
 msgid "break %s by ignoring some of its dependencies"
-msgstr "interrompi %s ignorando alcune delle sue dipendenze"
+msgstr "installa comunque %s ignorando alcune delle sue dipendenze"
 
 #: zypp/solver/detail/ProblemSolutionIgnore.cc:48
 msgid "generally ignore of some dependencies"


mlandres:
I closed bug#1173529 with reference to this issue. Next I'll check how we get it into WEBLATE, where the translations are maintained...

mlandres:
[WEBLATE](https://l10n.opensuse.org/projects/libzypp/master/it/#overview) should pick up the change from git in a while. If it does not, the change has to be applied there...

dfaggioli:
> [WEBLATE](https://l10n.opensuse.org/projects/libzypp/master/it/#overview) should pick up the change from git in a while. If it does not, the change has to be applied there...

Oh, ok, so I guess the change should have been made through that service? Sorry for my ignorance about that then! Please, do let me know if there is anything else/more I need to do for this very issue.

For future translations improvement, I'll check how WEBLATE works and use it directly.

And Thanks

mlandres:
No problem Dario. 

Right, WEBLATE usually pulls our .pot file and pushes back the .po files into git. 

I'm just curious whether it also works if I manip. and push a .po. I never tried. 
If WEBLATE does not pick it, I will apply your change there.

PR #241, opened 2020-08-28
Fixup VendorAttr API

mlandres:
Const-correct the VendorAttr API and make per target VendorAttr instances available in the Target API.

Rather than one global instance, now every target keeps it's on instance initialized with the data found below the targets root. Additionally there's one global ` VendorAttr::noTargetInstance`, which stores the settings used, if no Target is up (initialized with the settings below '/').

Todo:
- [x] Original VendorAttr Testcases  need to be fixed. They partially operate on the systems data and do not actually load the settings they want to load.
- [ ]  Implementation needs to be polished.


bzeller (approved):
OK once the missing rvalue ref is fixed ( as discussed in IRC )

PR #242, opened 2020-09-09
Readable testcases

bzeller:
Introduces a new testcase format that is more easy to read and write by humans. 

TODO: Still needs bump of libzypp version


zypp/misc/LoadTestcase.h
```diff
@@ -0,0 +1,130 @@
...
+#include <zypp/base/PtrTypes.h>
+#include <zypp/base/SetTracker.h>
+#include <zypp/base/NonCopyable.h>
+#include <zypp/sat/Queue.h>
+#include <zypp/target/modalias/Modalias.h>
+
+#include <optional>
+#include <memory>
+
+namespace zypp::misc {
```
mlandres, line 33: I would not use the `misc` namespace here. In case we extend the DefaultLoadSystem stuff, it may collide. Maybe zypp::misc::testcase or so.

mlandres (changes requested):
Except for the namespace it's

zypp/misc/LoadTestcase.h
bzeller: Done

mlandres (approved):

mlandres:
Merged manually (2605e947678fdde9641f3ccf035c80be36470329 - 186bf93e10582164c395da675fe7539434c83032)

PR #243, opened 2020-09-11
VendorAttr rework

mlandres:
The missing VendorAttr rewrite....

### DIFF
--- a/zypp/VendorAttr.cc
+++ b/zypp/VendorAttr.cc
@@ -107,7 +107,7 @@ namespace zypp
       if ( ! lcent )
       {
 	unsigned myid = 0;
-	// bnc#812608: no pefix compare in opensuse namespace
+	// bnc#812608: no prefix compare in opensuse namespace
 	static const IdString openSUSE( "opensuse" );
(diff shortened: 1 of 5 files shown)


zypp/VendorAttr.cc
```diff
@@ -99,86 +94,83 @@ namespace zypp
...
       }
     }
-	// add new entries
-    for_( it, vendorList_r.begin(), vendorList_r.end() )
-    {
-      _vendorMap[*it] = nextId;
+
+    // Now add the new entries
+    for ( std::string & vendor : vendorList_r ) {
+      if ( ! vendor.empty() )
```
bzeller, line 168: No need to go over the list if vendor is empty, so the if should enclose the for loop

bzeller (changes requested):

zypp/VendorAttr.cc
mlandres: You mixed `std::string & vendor` and `vendorList_r`, didn't you? We just skip empty strings in the list.

bzeller: Right I did not realize the vendor variable is shadowed here, sorry

bzeller (approved):

PR #246, opened 2020-09-23
PurgeKernels nagging..

mlandres:


zypp/PurgeKernels.cc
```diff
@@ -163,35 +162,35 @@ namespace zypp {
       return false;
     }
 
-    std::set<sat::Solvable::IdType> removedInThisRun;
-    removedInThisRun.insert( pi.id() );
+    std::set<sat::Solvable> removedInThisRun;
+    removedInThisRun.insert( slv );
 
-    for ( auto it = pool.byStatusBegin( toBeUninstalledFilter ); it != pool.byStatusEnd( toBeUninstalledFilter );  it++  ) {
+    for ( auto p : pool.byStatus( toBeUninstalledFilter ) ) {
```
bzeller, line 168: const auto & for range based?

zypp/PurgeKernels.cc
```diff
@@ -201,9 +200,8 @@ namespace zypp {
 
     //now check and mark the -debugsource and -debuginfo packages for this package and all the packages that were removed. Maybe collect it before and just remove here
     MIL << "Trying to remove debuginfo for: " << pi <<"."<<std::endl;
-    for ( const auto id : removedInThisRun ) {
+    for ( const auto solvable : removedInThisRun ) {
```
bzeller, line 203: same as above , const auto &

zypp/PurgeKernels.cc
mlandres: That's the downside of `auto`. For a `PoolItem` here `const&` makes sense. In the case below (and several other for loops) it's debatable whether a ` const sat::Solvable &` AKA `const unsigned &` is better than passing the unsigned by value.
Unfortunately `auto` disguises the types rationale. In my code you'd find `const PoolItem &` and `satSolvable` (not even const as it has no not-const methods).

bzeller (approved):
Ok

PR #250, opened 2020-10-07
Fix boost deprecated headers, remove dead code and obsolete #warning directives

mlandres:
Actually just ui/UserWantedPackages.cc is worth being reviewed to make sure I did not overlook something useful in the deleted code.


bzeller (approved):
Ok

PR #252, opened 2020-10-16
Fix networkmedia

bzeller:


tests/lib/WebServer.cc
```diff
@@ -309,11 +309,7 @@ class WebServer::Impl
...
           //connections and listen on a pipe for a wakeup event to stop the worker if required
-          int ret = ::poll( fds, 3, INT_MAX );
-          if ( ret == 0 ) {
-            //timeout just continue the loop condition will stop if required
-            continue;
-          }
+          int ret = ::poll( fds, 3, -1 );
 
           if ( ret < 0 ) {
             std::cerr << "Poll error " << endl;
```
mlandres, line 315: Even if just in tests/, if the error is logged, there should be the errno visible.
(btw bug#1177448: looks line YAST has some timer running firing every 10sec. We flood the log with EINTR errors. Harmless, but we should avoid this. Just in case you need another PR (for master) )

bzeller: Ok done

mlandres (approved):

PR #254, opened 2020-10-16
Handle eintr explicitely in EventDispatcher::waitForFdEvent

bzeller:


zypp/zyppng/base/eventdispatcher_glib.cc
```diff
@@ -455,6 +455,9 @@ bool EventDispatcher::waitForFdEvent( const int fd, int events , int &revents ,
         if ( timeout <= 0 )
           return false;
 
+        if ( errno == EINTR )
+          continue;
+
```
mlandres, line 460: Are we sure 'g_poll' sets 'errno' (and does not handle *EINTR` internally)?  And 'g_timer_elapsed' does not amend it?

Issue #257, opened 2020-10-26
proxy authentication: enable negotiate as additional authentication method


brassh:
At the moment, it seems that libzypp only allows proxy authentication via username/password, not via negotiate.

As far as I understand, proxy authentication for zypper works as follows:
1)
Proxy Server Address, port and transport are defined in the file /etc/sysconfig/proxy via variables, for example
HTTPS_PROXY="http://myproxy.domain:3128"
HTTPS_PROXY="http://myproxy.domain:3128" 
NO_PROXY="localhost,127.0.0.1,192.168.0.0/16"
2)
Proxy Username and Password are defined in the file /root/.curlrc via the line
--proxy-user="myusername:myverysecretpassword"
3)
If the user sets a systemwide proxy via YAST (yast/networking/proxy), YAST sets 1) and 2) above accordingly

This works, but leaving the proxy password in a non-encrypted file on the computer is a security risk (mitigated because the files in /root/.curlrc are readable only by root, but still a risk).

So the user should be given the choice to use negotiate authentication, if the proxy is kerberos enabled. 

I tried this for years, but it never worked (although curl supports negotiate). Problem seems that libzypp prevents curl from using it's negotiate capability. 

Today I found that it should be enough to change one single line in the source code to get negotiate authentication: 
Just add CURLAUTH_NEGOTIATE as additional auth method in file zypp/media/MediaCurl.cc, line 429

I tried it under openSUSE 15.2 (see, https://build.opensuse.org/package/show/home:brassh:branches:openSUSE:Leap:15.2:Update/libzypp) and it works for me.

With that simple change, the workflow could be is as follows:
1)
User enters only the proxy username in YAST, but leaves the proxy password empty. YAST sets username in /root/.curlrc accordingly, but without password) 
2)
Before using zypper, root user acquires credentials via "kinit myproxyusername"
3)
Now, "zypper ref", "zypper patch" etc. work flawlessly (at least on my box)
4)
Afterwards, root user can destroy the acquired credentials via "kdestroy". The password has never been stored on the box, neither encrypted nor unencrypted.

I do not know if there are negative side effects of the change, but did not find any yet. 

If this helps, I could submit a pull request (see, https://github.com/openSUSE/libzypp/compare/master...brassh:master)

PR #258, opened 2020-10-26
proxy authentication: enable negotiate as additional authentication method

brassh:
issue https://github.com/openSUSE/libzypp/issues/257

### DIFF
--- a/package/libzypp.changes
+++ b/package/libzypp.changes
@@ -1,4 +1,10 @@
 -------------------------------------------------------------------
+Mon Oct  26 18:43:12 CET 2020 - brassh@web.de
+
+- enable proxy authentication via negotiate
+- version 17.25.3 (22)
+
+-------------------------------------------------------------------
 Wed Oct  7 12:05:33 CEST 2020 - ma@suse.de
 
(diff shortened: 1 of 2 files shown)


bzeller:
Thanks for contributing but before merging this we will first need to make sure that we do not get failed requests in case that the server responds that negotiate is available but there are actually no credentials set by the user. 

The curl documentation says:
`If more than one bit is set, libcurl will first query the site to see what authentication methods it supports and then pick the best one you allow it to use`

Picking the auth is implemented here: https://github.com/curl/curl/blob/master/lib/http.c#L374  , which means if we signal that we want to use negotiate and the server supports it as well it would always be the first choice. 

I did not yet find code that would amend the list of auth methods we enable by checking if there are actually credentials available

brassh:
I see your point. This might break existing setups which rely on password based authentication.
A solution may be to activate negotiate only in case no password is given, see updated pull request.
In this case, the user can decide whether he wants password authentication (by providing proxyuser and password, as before) or password-less authentication (by providing only proxyuser, but no password in config or .curlrc)

brassh:
any comments?

PR #259, opened 2020-10-27
Minor fixes / Testcase r/w vendor lists

mlandres:


zypp/VendorAttr.cc
```diff
@@ -58,6 +58,24 @@ namespace zypp
...
+      for( const auto & el : _vendorGroupMap )
+	lists[el.second].push_back( el.first );
+
+      unsigned ret = 0;
+      for ( auto el : lists ) {
+	VendorList & vlist { el.second };
+	if ( vlist.empty() )
+	  continue;
+	++ret;
+	if ( fnc_r && !fnc_r( std::move(vlist) ) )
```
bzeller, line 73: Do you really want to move the vlist out of _vendorGroupMap? Since its a writeable reference into the map that would kill the data there..

zypp/misc/YamlTestcaseHelpers.h
```diff
@@ -197,6 +197,18 @@ namespace yamltest::detail {
...
         if ( !success ) return false;
       }
+      else if ( key == ("vendors") )
+      {
+        bool success = readListInlineOrFromFile( [&target]( const YAML::Node & dataNode, std::string * err ) {
+	  std::vector<std::string> vlist;
+	  for ( const auto & node : dataNode )
+	    vlist.push_back( node.as<std::string>() );
+	  if ( ! vlist.empty() )
+	    target.vendorLists.push_back( std::move(vlist) );
```
bzeller, line 207: Why do you first push into a std::vector<std::string> instead of directly pushing into target.vendorLists?

mlandres: Otherwise I had to remove empty vectors again from target.vendorLists (which is vector< vector<string> >)

zypp/VendorAttr.cc
mlandres: I move out of lists, not out of _vendorGroupMap.

bzeller: Right, having the variable "el" twice confused me

bzeller (approved):
OK

mlandres:
fixes  #255

PR #260, opened 2020-10-28
Protect ABI LoadTestcase types with a private implementation

bzeller:


zypp/misc/TestcaseSetup.h
```diff
@@ -0,0 +1,132 @@
...
+    ForceInstallImpl &data();
+  private:
+    RWCOW_pointer<ForceInstallImpl> _pimpl;
+  };
+
+  class TestcaseSetup
+  {
+  public:
+
+    TestcaseSetup();
```
mlandres, line 79: When using a pimpl I've got the habit to declare a DTOR (defined together with the CTOR, even if empty).
This prevents g++ from sometimes complaining about having to destroy the incomplete Impl class when a Class instance gets deleted (if the default dtor is used). Same applies to RepoData and ForceInstall.

mlandres (changes requested):

zypp/misc/TestcaseSetup.h
bzeller: Oh, yes very good point. I just forgot it because it did not complain this time, for whatever reason... --> Fixed

mlandres (approved):

PR #261, opened 2020-11-03
Ma 1050625 proxy credentials written to log files

mlandres:
1st draft stripping the proxpass except. within the .repo file.
And repo var replacement still needs to be checked.


bzeller (approved):
Ok

PR #267, opened 2020-11-25
bsc#1178910 fix rpm dbpath

mlandres:
[bsc#1178910](https://bugzilla.opensuse.org/show_bug.cgi?id=1178910)
Prefere an existing database, but follow rpms configured _dbpath if non exists. Assert a compat symlink at /var/lib/rpm exists in case the db is located elsewhere.

### DIFF
--- a/zypp/target/rpm/librpmDb.cc
+++ b/zypp/target/rpm/librpmDb.cc
@@ -96,15 +96,11 @@ public:
 //
 ///////////////////////////////////////////////////////////////////
 
-Pathname         librpmDb::_defaultRoot  ( "/" );
-// NOTE: Former variable, but now locked to "/var/lib/rpm".
-// A custom dbPath is not actually needed and would only work
-// reliably if libsolv also supports it.
-// The protected librpmDb ctor would allow to build a db_const_iterator
-// to access (ro) a database at a custom location.
-const Pathname   librpmDb::_defaultDbPath( "/var/lib/rpm" );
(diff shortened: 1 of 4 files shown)


zypp/target/rpm/librpmDb.cc
```diff
@@ -187,6 +179,30 @@ librpmDb * librpmDb::newLibrpmDb()
...
+  if ( ! root_r.absolute() )
+    ZYPP_THROW(RpmInvalidRootException( root_r, "" ));
+
+  // initialize librpm (for _rpmDefaultDbPath)
+  if ( ! globalInit() )
+    ZYPP_THROW(GlobalRpmInitException());
+
+  if ( PathInfo( root_r ).isDir() ) {
+    // If a known dbpath exsists, we continue to use it
+    for ( auto p : { "/var/lib/rpm", "/usr/lib/sysimage/rpm" } ) {
```
bzeller, line 194: Why do we go for the old default path first ? Is there a reason for it ?

bzeller (commented):
Good to go, but we have way too many statics. I get a bit of a headache with the _defaultDbPath and _rpmDefaultPath that get initialized in some other parts of the code. 
Best practice with statics is usually to initialize them right away or have accessors that do it lazily.

bzeller (approved):

mlandres:
> Good to go, but we have way too many statics. 

Welcome to legacy land.

PR #269, opened 2020-11-25
Update ZYppFactory.cc

medberry:
Fix typo.

Saw this in some recent zypper.log output and it is bothersome.


mlandres:
Thanks for the fix.

PR #272, opened 2020-12-03
bsc#1178910 libzypp wrong dbpath enhanced

mlandres:
[bsc#1178910](https://bugzilla.suse.com/show_bug.cgi?id=1178910)
Creating a librpmDb iterator before/without Target init finds the default dbpath unset and creates a database in /.

(The CheckAccessDeleted commit is unrelated. The numerous but superfluous log lines just caught my eye, because e.g. CheckAccessDeleted checks the installed lsof version and so triggers the above bug).

### DIFF
--- a/zypp/target/rpm/librpmDb.cc
+++ b/zypp/target/rpm/librpmDb.cc
@@ -125,6 +125,10 @@ bool librpmDb::globalInit()
   initialized = true; // Necessary to be able to use exand().
   _rpmDefaultDbPath = expand( "%{_dbpath}" );
 
+  if ( _rpmDefaultDbPath.empty() ) {
+    _rpmDefaultDbPath = "/usr/lib/sysimage/";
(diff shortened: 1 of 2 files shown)


zypp/target/rpm/librpmDb.cc
```diff
@@ -125,6 +125,10 @@ bool librpmDb::globalInit()
   initialized = true; // Necessary to be able to use exand().
   _rpmDefaultDbPath = expand( "%{_dbpath}" );
 
+  if ( _rpmDefaultDbPath.empty() ) {
+    _rpmDefaultDbPath = "/usr/lib/sysimage/";
```
bzeller, line 129: should we make that a  constexpr std::string_view somewhere ?

mlandres: IMO not needed yet. This is the only 'volatile' location. In case the distro changes the default again, it would be changed here. 
Down in suggestedDbPath the path occurs again, but in a list of known legacy paths where a DB could exist. But there the value is const and only more locations could be added.

bzeller (approved):

PR #275, opened 2020-12-10
Fix downloading metalink of small files

andrii-suse:
In case of small file download metalink file may be bigger than actual file.
MediaMurtiCurl would report an exception for such files:
'Downloaded data exceeded the expected filesize'.
This fix checks whether metalink flag is set and then sets hard limit of 16Kb for handling metalink response.

### DIFF
--- a/zypp/media/MediaMultiCurl.cc
+++ b/zypp/media/MediaMultiCurl.cc
@@ -1122,7 +1122,8 @@ multifetchrequest::run(std::vector<Url> &urllist)
 		}
 	    }
 
-	  if ( _filesize > 0 && _fetchedgoodsize > _filesize ) {
+	  const int maxMetalinkSize = 16*1024*1024;
+	  if ( _filesize > 0 && _fetchedgoodsize > _filesize && (!_context || !_context->_customHeadersMetalink || _fetchedgoodsize > maxMetalinkSize) ) {
 	    ZYPP_THROW(MediaFileSizeExceededException(_baseurl, _filesize));
 	  }
 	}


bzeller:
Did you actually run into this scenario? This code part actually is never executed if the Metalink file is still downloaded. 
The mutifetchrequest is only executed after the Metalink file was completely downloaded and parsed. 
What happens is:

-> Initial download with Metalink headers is executed without multifetch  [MediaMultiCurl.cc#L1356](https://github.com/openSUSE/libzypp/blob/master/zypp/media/MediaMultiCurl.cc#L1356)
-> After the initial file was downloaded the multi curl backend checks if its a metalink file [MediaMultiCurl.cc#L1390](https://github.com/openSUSE/libzypp/blob/master/zypp/media/MediaMultiCurl.cc#L1390)
-> If the backend decides its a Metalink file only then it would use the multifetch code where your change is: [MediaMultiCurl.cc#L1443](https://github.com/openSUSE/libzypp/blob/master/zypp/media/MediaMultiCurl.cc#L1443)

So if you get a exception at that point it should not be due to the Metalink file still being downloaded.

bzeller:
There is actually code that resets the expected filesize for the normal initial download if a Metalink is detected during download of the initial file data. This happens on the fly:

[MediaMultiCurl.cc#L1290](https://github.com/openSUSE/libzypp/blob/master/zypp/media/MediaMultiCurl.cc#L1290)

andrii-suse:
Yes, I have an automated test scenario, where the problem reliably happens on `zypper refresh`, which is reporting that size of https://mirrorcache.opensuse.org/download/distribution/leap/15.2/repo/oss/media.1/media exceeds 83 bytes. And size of metalink is about 8Kb.
To work around the problem in production I did patch mirrorcache code to not return metalink for "media" files.
Then I automated the flow, reproduced the problem and then have ensured that proposed patch fixes it.
Can it be that the problem happens very quickly, before size is reset?
Should I share the test case which sets up MirrorCache instance with few download.opensuse.org mirrors and then uses zypper against it? E.g. in CircleCI or on local machine (no root access is required) or in docker container?

bzeller:
If that fixes the problem it must be due to a different reason. If the code reaches that stage the MetaLink download is definitely over. Downloading the metalink and downloading the chunks does not use the same codepath.
We should investigate more why its happening. Do you have a mirrorcache instance where I can try to reproduce it locally against?

bzeller:
So I can locally reproduce the issue. 
The reason that this fails is that the Metalink files do not specify the filesize as they did before on Mirrorbrain. Seems the MediaMultiCurl backend does not expect that this could happen, tries to still download the file in chunks and uses the HTTP response headers content size as a indicator. 
But because its requesting chunks from the server it initializes it with a too small value.. I will need to debug a bit more to figure out what happens exactly

bzeller:
I will close this PR and create a issue for this instead. While the PR prevents the Exception from being thrown it is not correctly fixing the issue. 
Instead the MultiCurl backend will fall back to a simple download in the case when we do not have a already known filesize and do not receive one from the Metalink data.

PR #284, opened 2021-01-27
Minor logging issue regarding gpg keys

mlandres:
Mainly prevent any sichcheck error logging when we do not verify. In that case we just want to get the keys signing a message. The message file itself is /dev/null, so reporting 'bad signature' errors in the log is confusing.


bzeller (approved):
Ok, the codepath now never sets foundBadSignature but since we only use if if verify_r is also initialized it should be fine

PR #287, opened 2021-02-09
Workaround rpmdb2solv creating an empty db at /var/lib/rpm

mlandres:


bzeller (approved):
Ok, can be merged when CI build is green

PR #291, opened 2021-02-12
Fixup MediaHanderFactory / std::string_view in public API requires C++17

mlandres:


bzeller (approved):
Good, but fix the second commit message there is a typo:   MeadiaHandlerFactory  -> MediaHandlerFactory

mlandres:
fixed

PR #292, opened 2021-02-15
Patch: Identify well-known category names (bsc#117984)

mlandres:
The result of Patch::isCategory should be the same, no matter if
strings or enum values are compared.


bzeller (approved):
OK

PR #296, opened 2021-03-01
Media set verification issues

mlandres:
[bsc#1180851](https://bugzilla.suse.com/show_bug.cgi?id=1180851)
Reworked the MediaVerifier and relaxed the case where there is just one non-volatile mediun in the set.


bzeller (approved):
Ok

PR #297, opened 2021-03-05
Do no cleanup in custom cache dirs

mlandres:
[bsc#1182936](https://bugzilla.opensuse.org/show_bug.cgi?id=1182936)
Values read from the config file are remembered as default so RM can check whether a cachepath was later amended.


bzeller (approved):
ok

PR #303, opened 2021-03-18
Singletransrpm

bzeller:
This patch adds a alternative approach of handling multiple rpm operations in a single transaction via libzypp  and some required changes and addition to the existing code.


zypp/target/TargetImpl.cc
```diff
@@ -1541,12 +1543,39 @@ namespace zypp
 
       bool abort = false;
 
+      bool singleTrans = false;
```
bzeller, line 1546: Here its possible to switch between new and old path , we could use a env var for that

bzeller:
As discussed merged into DEV_NZYPP

PR #306, opened 2021-03-26
Fix unsafe usage of static in media verifier

andrii-suse:
Unless something changed dramatically recently in C++ - using `static` without `const` or mutex is always thread-unsafe.

### DIFF
--- a/zypp/repo/SUSEMediaVerifier.cc
+++ b/zypp/repo/SUSEMediaVerifier.cc
@@ -92,7 +92,7 @@ namespace zypp
 
       Pathname mediaFilePath( media::MediaNr mediaNr_r = 0 ) const
       {
-	static str::Format fmt { "/media.%d/media" };
+	str::Format fmt { "/media.%d/media" };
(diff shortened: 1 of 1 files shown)


bzeller:
This is thread safe since c++11 https://blog.mbedded.ninja/programming/languages/c-plus-plus/magic-statics/

bzeller:
I missed that he % operator changes the state of the format object, this is actually not a good idea. Reopening

andrii-suse:
> I missed that the % operator changes the state of the format object, this is actually not a good idea. Reopening

Yes, meanwhile I was troubleshooting zypper crashes when using mirrorcache - it crashes only on TW, but not on Leap, which indirectly confirms the problem. If you are interested, here is Dockerfile to reproduce:
```from opensuse/leap

RUN sed -i 's,http://download.opensuse.org,https://mirrorcache.opensuse.org,g' /etc/zypp/repos.d/*repo
RUN zypper --gpg-auto-import-keys ref
RUN zypper -vvv install -y -t pattern kde kde_plasma```

mlandres (commented):
yes

PR #309, opened 2021-04-26
Fix sevice detection with cgroupv2

mlandres:
Quick fix for [bsc#1184997](https://bugzilla.opensuse.org/show_bug.cgi?id=1184997)


bzeller (approved):
Ok

PR #310, opened 2021-04-28
Properly handle permission denied when providing optional files

mlandres:
[bsc#1185239](https://bugzilla.opensuse.org/show_bug.cgi?id=1185239)


bzeller (approved):
OK

PR #314, opened 2021-05-06
$ZYPP_LOCK_TIMEOUT: Let negative values wait forever for the zypp loc…

mlandres:
…k [bsc#1184399](https://bugzilla.suse.com/show_bug.cgi?id=1184399)

Helps boot time services like 'zypper purge-kernels' to wait for the zypp lock until other services using zypper have completed.

### DIFF
--- a/zypp/ZYppFactory.cc
+++ b/zypp/ZYppFactory.cc
@@ -17,7 +17,7 @@ extern "C"
 #include <fstream>
 #include <signal.h>
 
-#include <zypp/base/Logger.h>
+#include <zypp/base/LogControl.h>
 #include <zypp/base/Gettext.h>
 #include <zypp/base/IOStream.h>
 #include <zypp/base/Functional.h>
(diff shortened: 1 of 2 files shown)


zypp/ZYppFactory.cc
```diff
@@ -365,32 +365,41 @@ namespace zypp
...
-		MIL << "Waiting whether pid " << globalLock().lockerPid() << " ends within " << (LOCK_TIMEOUT-i) << " sec." << endl;
-                procdir = Pathname( Pathname("/proc")/str::numstring(globalLock().lockerPid()) );
-	      }
-	      else
-	      {
-		MIL << "Finally got the lock!" << endl;
-		break;	// gotcha
-	      }
 	    }
+	  } while ( failed && ( not giveup || Date::now() <= giveup ) );
```
Martchus, line 396: Is the use of alternative operators intentional? I personally don't have a strong opinion about them, I'm just wondering.

bzeller (approved):
Ok

Martchus (approved):
Thanks for taking care of bsc#1184399.

zypp/ZYppFactory.cc
mlandres: I think `not` is the only one I use sometimes. It may depend on the amount of python code I'm writing.

PR #322, opened 2021-06-12
Add comment option for lock command

coolgw:
Fix https://github.com/openSUSE/zypper/issues/388

### DIFF
--- a/zypp/PoolQuery.cc
+++ b/zypp/PoolQuery.cc
@@ -454,6 +454,9 @@ namespace zypp
 
     /** Kinds to search */
     Kinds _kinds;
+
+    /** comments */
+    std::string _comment;
     //@}
 
   public:
@@ -867,6 +870,9 @@ namespace zypp
   void PoolQuery::addKind(const ResKind & kind)
   { _pimpl->_kinds.insert(kind); }
 
+  void PoolQuery::setComment(const std::string & comment)
+  { _pimpl->_comment = comment; }
+
   void PoolQuery::addString(const std::string & value)
   { _pimpl->_strings.insert(value); }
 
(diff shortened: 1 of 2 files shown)


zypp/PoolQuery.cc
```diff
@@ -1387,6 +1402,9 @@ namespace zypp
           << it->idStr() << delim ;
     }
 
+    str << "comment: " << getComment() << delim ;
```
bzeller, line 1405: I think if you'd use the PoolQueryAttr::commentAttr here instead of the string it would be better

zypp/PoolQuery.cc
```diff
@@ -998,6 +1004,9 @@ namespace zypp
   PoolQuery::repos() const
   { return _pimpl->_repos; }
 
+  const std::string &
+  PoolQuery::getComment() const
+  { return _pimpl->_comment; }
```
bzeller, line 1009: The other getter functions don't use get prefix, please change it to use the same style.

zypp/PoolQuery.cc
bzeller: Also move the serializing of the comment to the end so that its the last element thats written to the file.

bzeller (changes requested):
A few small comments

zypp/PoolQuery.cc
coolgw: > I think if you'd use the PoolQueryAttr::commentAttr here instead of the string it would be better

fix now

coolgw: > Also move the serializing of the comment to the end so that its the last element thats written to the stream.

I have made some update, correct me if any misunderstanding

bzeller: You fixed the first part, but still need to move the code down to the end of the function. Currently you write the comment in the middle of all the other properties, but since its a new property I'd prefer to write it as the last element to the stream,
before the `str << delim;` line

coolgw: Fixed.

bzeller (approved):
Looks good to me, michael still needs to approve

coolgw:
@mlandres Could you help have a look?

mlandres (approved):
Sorry, for the delay. That's perfect.

PR #323, opened 2021-06-28
Fix bsc#1187738: purge-kernels fails with kernels from Kernel:HEAD

bzeller:
There recently was a change in the kernel package naming scheme in regards to rc kernels. 
Since kernel upstream uses characters in the version that are not allowed in rpm versions a "-rc" was previously replaced with ".rc" which broke sorting by  version, to fix this issue it was replaced with "~rc", which unfortunately broke the purge-kernels logic. 
This patch makes sure purge-kernel does apply the same conversion.

### DIFF
--- a/tests/zypp/PurgeKernels_test.cc
+++ b/tests/zypp/PurgeKernels_test.cc
@@ -328,6 +328,17 @@ namespace  {
           { "kernel-rt-devel-debuginfo-1-3.x86_64" },
         }
       },
+      // rc kernels have a special naming hack because upstream uses characters that are forbidden in rpm versions
+      TestSample {
+        TESTS_SRC_DIR"/zypp/data/PurgeKernels/rckrnl",
+        "5.13.0-rc7-17-default",
+        Arch("x86_64"),
+        "running",
+        {
+          { "kernel-default-5.13.rc7-16.1.x86_64" },
+          { "kernel-default-5.13~rc7-18.1.x86_64" }
+        }
+      },
     };
   }
 }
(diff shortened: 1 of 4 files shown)


hramrach:
The change to PurgeKernels.cc looks good, thanks.

Can't really say I understand the test code.

bzeller:
> The change to PurgeKernels.cc looks good, thanks.
> 
> Can't really say I understand the test code.

The test code change partially is just about better readability of the testcode and I added another testcase for the following case:
Installed kernels:
- "kernel-default-5.13~rc7-16.1.x86_64"
- "kernel-default-5.13~rc7-17.1.x86_64"
- "kernel-default-5.13~rc7-18.1.x86_64"

Running kernel uname:
`5.13.0-rc7-17-default`

Keepspec:
`running`

Expected removals:
- "kernel-default-5.13~rc7-16.1.x86_64"
- "kernel-default-5.13~rc7-18.1.x86_64"

mlandres:
> The change to PurgeKernels.cc looks good, thanks.
> 

@hramrach can you confirm, that the `-rc` kernels are used only internally? If customers have `-rc` kernels following the old naming schema installed, the *fixed* `purge-kernels` would fail for them, wouldn't it? In that case we would have to cover both cases.

hramrach:
The -rc kernels are not released, ever.

There is a development project where these are built and people can install them for testing.

However, they are not ever part of a Tumbleweed release, much less a stable distribution.

mlandres (approved):
lgtm

hramrach:
Is there something blocking this PR from merging?

PR #324, opened 2021-07-01
Show key fpr from signature when signature check fails (bsc#1187224)

bzeller:
Rpm by default only shows the short key ID when checking the signature
of a package fails. This patch reads the signatures from the RPM headers
and replaces she short IDs with the key fingerprints fetched from the signatures.


bzeller:
Reopened as https://github.com/openSUSE/libzypp/pull/327

PR #327, opened 2021-07-13
Show key fpr from signature when signature check fails (bsc#1187224) 

bzeller:
Rpm by default only shows the short key ID when checking the signature
of a package fails. This patch reads the signatures from the RPM headers
and replaces she short IDs with the key fingerprints fetched from the signatures.

### DIFF
--- a/zypp/target/rpm/RpmDb.cc
+++ b/zypp/target/rpm/RpmDb.cc
@@ -50,6 +50,7 @@ extern "C"
 #include <zypp/target/rpm/RpmException.h>
 #include <zypp/TmpPath.h>
 #include <zypp/KeyRing.h>
+#include <zypp/KeyManager.h>
 #include <zypp/ZYppFactory.h>
 #include <zypp/ZConfig.h>
 #include <zypp/base/IOTools.h>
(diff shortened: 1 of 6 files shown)


zypp/target/rpm/RpmDb.cc
```diff
@@ -1239,6 +1240,44 @@ namespace
...
+      addFprs( RPMTAG_SIGPGP );
+      addFprs( RPMTAG_RSAHEADER );
+      addFprs( RPMTAG_DSAHEADER );
+
+      for ( auto &detail : detail_r ) {
+        auto &line = detail.second;
+
+        // replace the short key ID with the long ones parsed from the header
+        for ( const auto &fpr : fprs ) {
+          str::replaceAll( line, fpr.first, fpr.second );
```
mlandres, line 1276: As Michael said, rpm may report the full FPR in the future. 
Reading the FPR from the header and replacing the last 8 byte of the reported ID (L1261) must not happen if the reported id is longer than 8.

mlandres (changes requested):

zypp/target/rpm/RpmDb.cc
bzeller: Done

zypp/target/rpm/RpmDb.cc
```diff
@@ -1239,6 +1240,54 @@ namespace
...
+
+          if ( !didReadHeader ) {
+            didReadHeader = true;
+
+            // Get signature info from the package header, RPM always prints only the 8 byte ID
+            auto header = RpmHeader::readPackage( path_r, RpmHeader::NOVERIFY );
+            if ( !header ) {
+              ERR << "Failed to read package" << std::endl;
+              return RpmDb::CHK_ERROR;
+            }
```
mlandres, line 1262: If the header is for whatever reason not readable, I'd prefer to continue with the original workflow without replacing the FPR. Just leave the loop and return the computed value (ret).

mlandres (changes requested):

zypp/target/rpm/RpmDb.cc
bzeller: ok

mlandres (approved):

PR #331, opened 2021-07-26
Rephrase vendor conflict message in case 2 packages are involved (bsc…

mlandres:
…#1187760)

This covers the case where not the packages itself would change
its vendor, but replaces a package from a different vendor.

### DIFF
--- a/zypp/solver/detail/SATResolver.cc
+++ b/zypp/solver/detail/SATResolver.cc
@@ -1489,7 +1489,17 @@ SATResolver::problems ()
 				{
                                     IdString s_vendor( s.vendor() );
                                     IdString sd_vendor( sd.vendor() );
-				    std::string description = str::Format(_("install %1% (with vendor change)\n  %2%  -->  %3%") ) % sd.asString() % ( s_vendor ? s_vendor.c_str() : " (no vendor) " ) % ( sd_vendor ? sd_vendor.c_str() : " (no vendor) " );
+				    std::string description;
+				    if ( s == sd )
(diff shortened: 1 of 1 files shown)


zypp/solver/detail/SATResolver.cc
```diff
@@ -1489,7 +1489,17 @@ SATResolver::problems ()
 				{
                                     IdString s_vendor( s.vendor() );
                                     IdString sd_vendor( sd.vendor() );
-				    std::string description = str::Format(_("install %1% (with vendor change)\n  %2%  -->  %3%") ) % sd.asString() % ( s_vendor ? s_vendor.c_str() : " (no vendor) " ) % ( sd_vendor ? sd_vendor.c_str() : " (no vendor) " );
+				    std::string description;
+				    if ( s == sd )
+				      description = str::Format(_("install %1% (with vendor change)\n  %2%  -->  %3%") )
```
bzeller, line 1494: "install" > "installing"

zypp/solver/detail/SATResolver.cc
```diff
@@ -1489,7 +1489,17 @@ SATResolver::problems ()
...
+				      % sd.asString()
+				      % ( s_vendor ? s_vendor.c_str() : " (no vendor) " )
+				      % ( sd_vendor ? sd_vendor.c_str() : " (no vendor) " );
+				    else
+				      description = str::Format(_("install %1% from vendor %2%\n  replacing %3% from vendor %4%") )
```
bzeller, line 1499: same as above

bzeller (changes requested):

bzeller (approved):

PR #339, opened 2021-08-27
Fix issues in medianetwork

bzeller:
This patch fixes some issues in the downloader and medianetwork code.
Especially with recursive calling of CURL API which is always a bad idea


mlandres (approved):
Good. I'll merge it.

PR #341, opened 2021-09-02
Do not double check signatures and keys (bsc#1190059)

bzeller:
Since we already check signatures and keys in libzypp there is no need
to check them again in the zypp-rpm backend. This patch adds a new
install flag and disables redundant checks in zypp-rpm.

### DIFF
--- a/tools/zypp-rpm/main.cc
+++ b/tools/zypp-rpm/main.cc
@@ -224,21 +224,22 @@ int main( int, char ** )
   int tsFlags           = RPMTRANS_FLAG_NONE;
   int tsVerifyFlags     = RPMVSF_DEFAULT;
 
-  if ( msg.flags() & RpmInstFlag::RPMINST_NODIGEST)
+  const auto &rpmInstFlags = msg.flags();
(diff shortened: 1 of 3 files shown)


tools/zypp-rpm/main.cc
```diff
@@ -270,6 +271,8 @@ int main( int, char ** )
   // the transaction data we will get in the callback
   TransactionData data { msg };
 
+  // do we care about knowing the public key?
```
mlandres, line 274:
```suggestion
  // We may not need the public key if the package is secured by a strong checksum in signed metadata.
```

mlandres (approved):
Except for the comment it's fine.

tools/zypp-rpm/main.cc
bzeller: This is for the libzypp side to decide, in zypp-rpm all we care about is if that switch is set or not. I'd suggest to move that comment to TargetImpl.cc instead.

bzeller: ( maybe with reference here like: // see TargetImpl.cc : line )

PR #344, opened 2021-09-08
Make sure to keep states alives while transitioning (bsc#1190199)

bzeller:
The statemachine needs to make sure a state is not deleted while enter()
was called. This patch should make sure that the states live long
enough.


mlandres (approved):
ok

PR #347, opened 2021-09-23
Fix kernel-*-livepatch removal in purge-kernels

lpechacek:
This change is required for the smooth integration of SLE Live Patching into SLE Micro. The REGEX for kernel-*-livepatch matching might perhaps be improved but I believe that this simple one is enough.


bzeller (approved):
Looks good to me, but we should add a testcase for the specific livepatch packages

PR #348, opened 2021-10-05
RepoManager: Don't probe for plaindir repo if URL schema is plugin

mlandres:
[bsc#1191286](https://bugzilla.suse.com/show_bug.cgi?id=1191286)


bzeller (approved):
Ok

PR #349, opened 2021-10-07
Use procfs to detect nr of open fd's if rlimit is too high (bsc#1191324)

bzeller:
Especially in a VM iterating over all possible fd's to close open ones
right before a exec() slows down zypper unnecessarily. This patch uses
/proc/self/fd to iterate over open fd's in case rlimit is above 1024.

### DIFF
--- a/zypp-core/fs/PathInfo.cc
+++ b/zypp-core/fs/PathInfo.cc
@@ -538,35 +538,64 @@ namespace zypp
     }
 
     ///////////////////////////////////////////////////////////////////////
-    // dirForEach
+    // dirForEachImpl
     ///////////////////////////////////////////////////////////////////////
+    template <class... T>
+    constexpr bool always_false = false;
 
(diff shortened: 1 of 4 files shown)


crrodriguez:
huh...no.. if you want close all open fds you have to use the close_range() system call and only do this /proc/self/fd iteration if invoking syscall return negative value..

mlandres:
Yes, we could check whether close_range is available. AFAICS not up to 15.3. Don't know if it's in 15.4, but TW seems to have it.

crrodriguez:
> Yes, we could check whether close_range is available. AFAICS not up to 15.3. Don't know if it's in 15.4, but TW seems to have it.

even better, it should just use posix_spawn to execute things if the file action to close file descriptors on exec is available in the C library.. if posix_spawn_file_actions_addclosefrom_np is defined.. 
It is not that flexible but it is the cheapest/fastest way to run commands

PR #351, opened 2021-10-13
Require a minimum number of mirrors for multicurl (bsc#1191609)

bzeller:


mlschroe:
This is wrong, it kills delta downloads of repository metadata. It should only do this if there is no blocklist or the blocklist does not contain holes.

PR #354, opened 2021-10-31
WIP: Store key signatures on public keys

dwymark:
Add a new struct to hold data for signatures on public keys, and use this struct to store signature data on every PublicKeyData. Provide functions to determine whether a signature is trusted or known.

This work is being done with [zypper issue 348](https://github.com/openSUSE/zypper/issues/348) in mind.

### DIFF
--- a/zypp/PublicKey.cc
+++ b/zypp/PublicKey.cc
@@ -214,12 +214,6 @@ namespace zypp
   public:
     /** Offer default Impl. */
     static shared_ptr<Impl> nullimpl();
-    static Target & target(const Pathname & sysRoot = "/")
-    {
-      if ( ! getZYpp()->getTarget() )
-        getZYpp()->initializeTarget( sysRoot );
-      return *getZYpp()->getTarget();
-    }
 
   private:
     friend Impl * rwcowClone<Impl>( const Impl * rhs );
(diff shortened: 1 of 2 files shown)


zypp/PublicKey.cc
```diff
@@ -196,6 +199,114 @@ namespace zypp
...
+    }
+    return str::Str() << nameStr
+                      << id() << " "
+                      << created().printDate()
+                      << " [" << expiresDetail( expires() ) << "]";
+  }
+
+  bool PublicKeySignatureData::inTrustedRing() const
+  {
+    std::list<PublicKey> allKeys = _pimpl->target().rpmDb().pubkeys();
```
dwymark, line 299: @mlandres Do you know if this is an appropriate method for accessing the public keys in the rpmdb? Here is the definition of `target` for ease of reference:

```
    static Target & target(const Pathname & sysRoot = "/")
    {
      if ( ! getZYpp()->getTarget() )
        getZYpp()->initializeTarget( sysRoot );
      return *getZYpp()->getTarget();
    }
```

 This implementation is based on code that I saw in some unit tests to get the current target. I am not sure if it is appropriate to use this same logic in the main code.

mlandres: The Target access here is not the right way. Depending on the workflow, we may have to handle keys and repos for the software selection before the target system is ready and mounted (e.g. for bare metal installation). Authority is always our KeyRing and it's synced behind the scenes with the rpmdb.

[KeyRing::isKeyTrusted](https://github.com/openSUSE/libzypp/blob/master/zypp/KeyRing.h#L259) and [KeyRing::isKeyKnown](https://github.com/openSUSE/libzypp/blob/master/zypp/KeyRing.h#L265) should work here; accessed via ` getZYpp()->keyRing()->`.

PR #357, opened 2021-11-08
Disable logger in the child after fork (bsc#1192436)

bzeller:


Vogtinator:
Why not use `pthread_atfork`?

zypp-core/zyppng/io/forkspawnengine.cc
```diff
@@ -103,7 +104,9 @@ void zyppng::AbstractDirectSpawnEngine::mapExtraFds ( int controlFd )
   const auto maxFds = ( ::getdtablesize() - 1 );
   //If the rlimits are too high we need to use a different approach
   // in detecting how many fds we need to close, or otherwise we are too slow (bsc#1191324)
-  if ( maxFds >= 1024 && zypp::PathInfo( "/proc/self/fd" ).isExist() ) {
+  std::vector<int> fdsToClose;
```
Vogtinator, line 107: Only used within the `if`, scope can be reduced

bzeller:
> `pthread_atfork`

Good point, will change it

zypp-core/zyppng/io/forkspawnengine.cc
bzeller: Good point

PR #361, opened 2021-11-27
Some enhancements around reporting (traditional rpm and singletrans)

mlandres:
A few unrelated commits at the beginning.

Enhanced rpm script output reports to offer output lines immediately. Zypper will be enhanced to show them immediately rather than showing the collected output at the end.

For singetrans a new report is used to propagate the (debug)lines issued from the rpmlog subsystem.  Zypper will be enhanced to show them in addition to sny script output.

### DIFF
--- a/zypp/target/TargetImpl.cc
+++ b/zypp/target/TargetImpl.cc
@@ -1896,7 +1896,8 @@ namespace zypp
 
       bool abort = false;
       zypp::AutoDispose<std::unordered_map<int, ManagedFile>> locCache([]( std::unordered_map<int, ManagedFile> &data ){
-        for ( auto &[key, value] : data ) {
+        for ( auto &[_, value] : data ) {
+          (void)_; // unsused; for older g++ versions
           value.resetDispose();
         }
         data.clear();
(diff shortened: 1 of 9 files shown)


zypp/target/TargetImpl.cc
```diff
@@ -1896,7 +1948,8 @@ namespace zypp
 
       bool abort = false;
       zypp::AutoDispose<std::unordered_map<int, ManagedFile>> locCache([]( std::unordered_map<int, ManagedFile> &data ){
-        for ( auto &[key, value] : data ) {
+        for ( auto &[_, value] : data ) {
+          [[maybe_unused]] auto & key { _ };
```
bzeller, line 1952: Does that silence a warning?  It seems GCC8 can correctly handle the [[maybe_unused]] directly in the for() declaration

bzeller: I guess the optimizer will remove taking the reference there, but it still looks not so nice

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wunused-variable"
    for (auto & [unused, val]: my_map)
#pragma GCC diagnostic pop
```

also is possible but is not exactly pretty

bzeller (changes requested):

zypp/target/TargetImpl.cc
mlandres: Using `(void)_;` to silence g++ < 8

PR #363, opened 2021-12-12
Handle armv8l as armv7hl compatible userland

dirkmueller:
When invoking 32bit mode in userland of an aarch64 kernel the
kernel identifies itself as 'armv8l'. This is a special mode
that is minimally userland compatible with 'armv7hl' (however not
with other versions of the armvX ABI). Map this according to
how rpm(1) handles in libzypp so that zypper can be used as well.


dirkmueller:
(tested via https://build.opensuse.org/project/show/home:dirkmueller:branches:openSUSE:Factory:ARM )

ggardet:
LGTM. Thanks for the link to the tests in OBS.

mlandres:
Thanks!

zypp/parser/yum/schema/common-inc.rnc
```diff
@@ -52,6 +52,7 @@ private.archenum = "noarch"
             | "sparcv8"
             | "sparc"
             | "aarch64"
+            | "armv8l"
```
Vogtinator, line 55: Missing `armv8hl`

zypp/parser/yum/schema/common-inc.rng
```diff
@@ -115,6 +115,7 @@
       <value>sparcv8</value>
       <value>sparc</value>
       <value>aarch64</value>
+      <value>armv8l</value>
```
Vogtinator, line 118: Missing `armv8hl`

Vogtinator:
The relevant RPM change is https://github.com/rpm-software-management/rpm/commit/b1f81b837f46bcbad8b527e41c7b7ebb0175d156

mlandres:
Fixed schema files: d5e132ad3288d4f8f4cb78bf0ab41adf510bbfdf

PR #364, opened 2021-12-15
Fix wrong encoding of iso: URL components (bsc#954813)

mlandres:
There is no need to fiddle with the encoding. If the iso: URL has a
url= embedded in the query part, the retrieved value must be a valid
- special chars encoded - URL. And if the retrieved iso= filename
contains special chars (spaces,etc.), Url::setPathName will properly
encode them. There should be no need for any manual url::encode call.


bzeller (approved):
ok

PR #374, opened 2022-01-30
Fix exception handling when reading or writing credentials (bsc#1194898)

mlandres:
https://bugzilla.suse.com/show_bug.cgi?id=1194898


bzeller (approved):
ok

PR #387, opened 2022-03-24
Wrong solver settings when switching the target

mlandres:
[bsc#1196368](https://bugzilla.suse.com/show_bug.cgi?id=1196368)


zypp/solver/detail/Resolver.cc
```diff
@@ -100,15 +112,42 @@ Resolver::~Resolver()
...
 
+
+void Resolver::setDefaultSolverFlags( bool all_r )
+{
+  MIL << "setDefaultSolverFlags all=" << all_r << endl;
+
+  if ( all_r || _applyDefault_focus ) setFocus( ResolverFocus::Default );
+
+#define ZOLV_FLAG_DEFAULT( ISDEFAULT, ZSETTER )        \
+  if ( all_r || ISDEFAULT ) ZSETTER( indeterminate )
```
bzeller, line 123: Using template functions are cleaner than defines. We should consider using those if possible

bzeller (commented):
Looks good aside from a few small comments. We also could think about a testcase for this

zypp/ZConfig.cc
```diff
@@ -812,10 +845,12 @@ namespace zypp
   ZConfig::~ZConfig( )
   {}
 
+  void ZConfig::sigTargetChanged()
```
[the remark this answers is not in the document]
mlandres: Yes, I'll change it.

zypp/solver/detail/Resolver.cc
mlandres: This way It's in sync with the setter/getter defines. It strengthens the desire that all these bit-Flags should follow the same schema.  I would agree if such a macro was used at various locations throughout the code, but it's a continuous block and IMO leaves litte room for misuse.
```
#define ZOLV_FLAG_DEFAULT( ISDEFAULT, ZSETTER )        \
  if ( all_r || ISDEFAULT ) ZSETTER( indeterminate )

  ZOLV_FLAG_DEFAULT( _applyDefault_forceResolve,         setForceResolve         );
  ...
  ZOLV_FLAG_DEFAULT( _applyDefault_dupAllowVendorChange, dupSetAllowVendorChange );
#undef ZOLV_FLAG_TRIBOOL
```

PR #390, opened 2022-04-21
Work around cases where sat repo.start points to an invalid solvable

mlandres:
May happen if (wrong arch) solvables were removed at the beginning
of the repo.

### DIFF
--- a/zypp/sat/Solvable.h
+++ b/zypp/sat/Solvable.h
@@ -501,10 +501,11 @@ namespace zypp
     namespace detail
     {
       ///////////////////////////////////////////////////////////////////
-      //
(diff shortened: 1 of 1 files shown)


bzeller (changes requested):
This only adjusts the start iterator, however in my case start and end are 0 . Or would that work correctly?

mlandres:
Moved the fix to the SolvableIterator itself. It already iterates only over valid solvables, so if the initial Solvable is not valid, the Iterator now advances to the next valid one (or noSolvable at the end of the pool) in the ctor.

bzeller (approved):
Ok

PR #395, opened 2022-05-19
Immediately start monitoring the download.transfer_timeout 

mlandres:
[bsc#1199042](https://bugzilla.opensuse.org/show_bug.cgi?id=1199042)
Do not wait until the first data arrived.


bzeller (approved):
Ok

PR #416, opened 2022-10-02
Don't do 2 HTTP requests when one is enough

bmwiedemann:
improves time of zypper ref from 10.7 to 9.4s
and even more on high-latency connections

TODO: test handling if the file does not exist


mlandres:
Basically yes, we just have to fix the progress reporting. Customers complained about _errors_ being reported in verbose mode
```
Retrieving: repomd.xml ...........................[done]
Retrieving: media ...........................[not found]
```

PR #419, opened 2022-10-07
Skip media.1/media download for http repo status calc

bzeller:
This patch allows zypp to skip a extra media.1/media download to calculate if a repository needs to be refreshed. This optimisation only takes place if the repo does specify only downloading base urls.

### DIFF
--- a/zypp/RepoInfo.cc
+++ b/zypp/RepoInfo.cc
@@ -27,6 +27,7 @@
 #include <zypp/Pathname.h>
 #include <zypp/ZConfig.h>
 #include <zypp/repo/RepoMirrorList.h>
+#include <zypp/repo/SUSEMediaVerifier.h>
 #include <zypp/ExternalProgram.h>
 
 #include <zypp/base/IOStream.h>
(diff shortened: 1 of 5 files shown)


zypp/RepoInfo.cc
```diff
@@ -1047,6 +1048,26 @@ namespace zypp
     return str << "GpgCheck::UNKNOWN";
   }
 
+  bool RepoInfo::requireStatusWithMediaFile () const
+  {
+    bool canSkipMediaCheck = std::all_of( baseUrlsBegin(), baseUrlsEnd(), []( const zypp::Url &url ) { return url.schemeIsDownloading(); });
+
+    const auto &mDataPath = metadataPath();
+    if ( canSkipMediaCheck && !mDataPath.empty() ) {
+      zypp::Pathname mediafile = mDataPath/"media.1"/"media";
```
mlandres, line 1057:
```suggestion
      zypp::Pathname mediafile = mDataPath/"media.1/media";
```

zypp/RepoInfo.cc
```diff
@@ -1047,6 +1048,26 @@ namespace zypp
...
+
+      DBG << "Trying mediafile from: " << mediafile << std::endl;
```
mlandres, line 1059: The line is IMO not needed. SUSEMediaVerifier logs an error if the file can't be parsed and the DBG at the end contains all info needed,

zypp/RepoInfo.h
```diff
@@ -521,6 +521,8 @@ namespace zypp
       LocaleSet getLicenseLocales( const std::string & name_r ) const;
      //@}
 
+      bool requireStatusWithMediaFile () const;
```
mlandres, line 524: `/** what is is doing. */`

mlandres (changes requested):

zypp/RepoInfo.cc
bzeller: OK

zypp/RepoInfo.cc
bzeller: Done

zypp/RepoInfo.h
bzeller: Done

mlandres (approved):

PR #422, opened 2022-10-11
Implement geoIP feature for zypp

bzeller:
This patch adds a feature to rewrite request URLs to the repo servers by querying a geoIP file from download.opensuse.org. This file can return a redirection target depending on the clients IP adress, this way we can directly contact a local mirror of d.o.o instead. The redir target stays valid for 24hrs.


zypp/media/MediaNetworkCommonHandler.cc
```diff
@@ -0,0 +1,81 @@
...
+      }
+    }
+
+    // no rewrite
+    return Url();
+  }
+
+  Url MediaNetworkCommonHandler::getFileUrl( const Pathname & filename_r ) const
+  {
+    constexpr std::string_view invalidRewrites[] = { "repomd.xml.asc", "repomd.xml.key", "geoip", std::string_view() };
```
andrii-suse, line 60: We should add repomd.xml here as well.
Regional MirrorCache instance does not try to guarantee that its repomd.xml is the latest (it is tricky to guarantee it anyway), so it will redirect requests to repomd.xml tho the main instance.
So it would be better if zypper doesn't use geoip for it as well.
Also media.1/media is tricky file - I am not sure if zypper will still try to access it, but better do not use geoip for it as well (because it is tricky to guarantee freshness of non-versioned files).

andrii-suse (commented):
Overall looks good as far as I understand, just the comment about `invalidRewrites` array and maybe we need a new option in config file to force disable such behavior (e.g. when geoip database is incorrect and the client gets redirected to a wrong region).

zypp/media/MediaNetworkCommonHandler.cc
bzeller: Done

mlandres (approved):

PR #425, opened 2022-10-21
Properly reset range requests

mlandres:
[bsc#1204548](https://bugzilla.suse.com/show_bug.cgi?id=1204548)

Stupid C error. All the error handling locations properly reset the head/range request curl settings. But the most important case, when the task succeeded, failed to reset the range request. Subsequent tasks using the same curl handle continue to request just 2 byte. Fortunately this happens only if `?head_requests=no` is set in the URL.

Now using a RAII guard to make sure the options are reset no matter how the function is left.


bzeller (approved):
OK, note that we also have zypp::OnScopeExit if you need to defer cleanups to the end of the current scope.

PR #426, opened 2022-11-03
'.no_auto_prune' in the package cache dir prevents auto cleanup

mlandres:
[bsc#1204956](https://bugzilla.opensuse.org/show_bug.cgi?id=1204956)

In conjunction with `--keep-packages` set for the repositories, downloaded packages will stay in the package cache, even if repositories are removed or no longer present on the system. Explicit `zypper clean` is not affected (will clean the packages).

If this is actually desired: `touch /var/cache/zypp/packages/.no_auto_prune`
(for libzypp-17.31.5)

### DIFF
--- a/zypp/RepoManager.cc
+++ b/zypp/RepoManager.cc
@@ -455,6 +455,10 @@ namespace zypp
     };
     ////////////////////////////////////////////////////////////////////////////
 
+    /** bsc#1204956: Tweak to prevent auto pruning package caches. */
+    inline bool autoPruneInDir( const Pathname & path_r )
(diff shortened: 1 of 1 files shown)


bzeller (changes requested):
Can we document the new file in the man pages? Otherwise there is no way people can discover this feature

bzeller (approved):
OK

kuanyui:
Seeking for this feature for a long time. Just out of curiosity, may be a dumb question: why is this feature not in `/etc/zypp/zypp.conf` instead?

PR #437, opened 2023-01-17
Extended tests

bzeller:


mlandres:
We'll need additional buildrequires in the .spec `%if %{with testrunner}`?

bzeller:
> We'll need additional buildrequires in the .spec `%if %{with testrunner}`?

The testrunner itself does not really have extra dependencies... Even if testrunner is enabled the tests are not executed during build and I don't really see the possibility to do so.

PR #443, opened 2023-03-11
capability parse richdeps

mlandres:
Capability legacy is to parse everything unknown into a NAMED cap. Parsing from string must not return Capability::Null.

But more important: fix CapDetail to return Rel::NONE if an EXPRESSION is used as a NAMED cap.
This bug was disguised by the fact that we did not  parse richdeps (or other EXPRESSIONs) from user supplied strings before. Wrongly treating an EXPRESSION as NAMED for a query is a bug per se, but  with this fix it should at least match nothing, rather than everything.


bzeller (approved):
ok

PR #445, opened 2023-03-23
Teach MediaNetwork to retry on HTTP2 errors

bzeller:


mlandres (approved):

crrodriguez:
Im seeing this errores constantly..  patch looks good..but you shall handle CURLE_HTTP3 #if defined() the same way.

bzeller:
> Hello..Im seeing this errors constantly.. patch looks good..but you shall handle CURLE_HTTP3 #if defined() the same way.

Thanks for the hint but HTTP/3 is currently not supported and used by zypper. This explicit handling is only required because our server infrastructure gets overloaded after big distribution rebuilds and we need to auto retry on those errors. Once we switch to HTTP/3 and get similar issues there we will add the error codes if required.

PR #446, opened 2023-03-24
multicurl: propagate ssl settings stored in repo url

mlandres:
[bsc#1127591](https://bugzilla.opensuse.org/show_bug.cgi?id=1127591)
fixes #335

### DIFF
--- a/zypp-curl/parser/metalinkparser.cc
+++ b/zypp-curl/parser/metalinkparser.cc
@@ -23,6 +23,19 @@
 
 using namespace zypp::base;
 
+namespace zypp::env
+{
+  /** Hack to circumvent the currently poor --root support. */
+  inline bool ZYPP_METALINK_DEBUG()
+  {
+    static bool val = [](){
+      const char * env = getenv("ZYPP_METALINK_DEBUG");
+      return( env && zypp::str::strToBool( env, true ) );
+    }();
+    return val;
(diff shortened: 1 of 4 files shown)


zypp-curl/parser/metalinkparser.cc
```diff
@@ -431,6 +442,11 @@ MetaLinkParser::parse(const InputStream &is)
       parseBytes(buf, is.stream().gcount());
     }
   parseEnd();
+  MIL << "Parsed " << pd->urls.size() << " mirrors from " << is.path() << std::endl;
+  if ( env::ZYPP_METALINK_DEBUG() ) {
+    for ( const auto &mirr : pd->urls )
+      DBG << "- " << mirr.priority << " " << mirr.url << std::endl;
+  }
```
bzeller, line 449: Should we add the same in libzypp/zypp-curl/ng/network/private/downloaderstates/preparemulti_p.cc:122 ?

mlandres: The parser itself issues the messages. As long as you didn't implement your own MetaLinkParser it shoudl do.

bzeller: Then we also should add it to the ZSyncParser

bzeller (approved):

PR #459, opened 2023-05-24
Do not unconditionally release a medium if provideFile failed (bsc#12…

mlandres:
[bsc#1211661](https://bugzilla.suse.com/show_bug.cgi?id=1211661)

MediaSetAccess continues and re-uses an attached medium as long as files are provided successfully. Unconditionally releasing the medium after a file could not be provided is not needed. We can continue with the next file, if there is one.

We can leave it to the exception handling. If the MediaSetAccess goes out of scope, it's released anyway. If the exception is handled earlier, we can continue to use it.

Only in case we interactively request a new medium from the user, the medium needs to be discarded. But this case calls MediaManager::releaseAll anyway.

### DIFF
--- a/zypp/MediaSetAccess.cc
+++ b/zypp/MediaSetAccess.cc
@@ -293,20 +293,6 @@ IMPL_PTR_TYPE(MediaSetAccess);
 
         do
         {
-          if (user != media::MediaChangeReport::EJECT) // no use in calling this again
-          {
-            DBG << "Media couldn't provide file " << file << " , releasing." << endl;
-            try
-            {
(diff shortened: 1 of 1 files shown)


mlandres:
For bsc#1211661 we are interested in fixing the case where a mandatory file (repomd.xml) is downloaded successfully. 

While checking the file, we may parse it and download additional optional keys on the fly. We must not discard everything - especially not the repomd.xml itself - if one of them fails. If the repomd.xml is validated successfully, everything is moved to the cache.

bzeller (approved):

PR #463, opened 2023-06-14
build: honor libproxy.pc's includedir

DimStar77:
Fixes build against libproxy 0.5, where the proxy.h is moved to /usr/include/libproxy

the .pc file properly advertises this though

Addresses the failure seen in https://build.opensuse.org/package/live_build_log/home:dimstar:libproxy/libzypp/openSUSE_Factory/x86_64


mlandres:
Thank you.

PR #465, opened 2023-06-27
Fix build against protobuf >= 22

Vogtinator:
Protobuf 22 has some changes which broke using the CMake supplied FindProtobuf.cmake (see https://gitlab.kitware.com/cmake/cmake/-/issues/24321 and https://github.com/protocolbuffers/protobuf/issues/12292).

"Port away from protobuf_generate_cpp" migrates away from a (apparently) deprecated function to using `protobuf_generate` which works with both the cmake and protobuf supplied module. "Have zypp-proto depend on Protobuf-Lite via the imported target" makes it such that protobuf is only found and used in a single place, making the final switch to the protobuf supplied cmake module in the third commit easier.

Locally it builds against protobuf 21.12 (currently in TW) and 22.5 (currently in Staging:K).


Vogtinator:
Fails on SP1, meh:

```
[  120s] CMake Error at zypp-proto/CMakeLists.txt:5 (protobuf_generate):
[  120s]   Unknown CMake command "protobuf_generate".
```

That means there is no common ground between the cmake module interface in SP1 and the protobuf module interface.

Not sure what to do here.

Vogtinator:
Doing it like https://github.com/protocolbuffers/protobuf/blob/3a95ea12cd42b4d23e794803db5cc71901797e18/examples/CMakeLists.txt#L28 and using `protobuf_generate_cpp` if `protobuf_generate` is not available doesn't directly work, because the generated header files are not in subdirectories anymore. Sigh.

dirkmueller:
Looks like the problematic distro failure is 15.1 as that is still in LTSS (according to https://www.suse.com/lifecycle#suse-linux-enterprise-server-15 ). 15.0 could be removed from testing.

dirkmueller:
Thanks for the porting effort. I filed #466 as a stop gap so that tumbleweed can continue to roll..

Vogtinator:
I was able to make it happy by importing `FindProtobuf.cmake` from cmake into this repo. Not pretty, but the least bad option I found.

mlandres:
Thanks @Vogtinator

PR #466, opened 2023-06-27
Limit to protobuf < 22

dirkmueller:
Protobuf 22 comes with a number of breaking changes, see https://protobuf.dev/support/migration/#cpp-22


mlandres:
Due to #465 this PR is not needed.

PR #473, opened 2023-07-29
Revert "Color download progress bar"

jengelh:
This reverts the code originally added in zypper's commit f141cf5579a460cb46093dfa130f25b03efaefce and which was later moved to libzypp.

Cyan is already used for the output of RPM scriptlets. Avoid this colorific collision between download progress bar and scriptlet output. See
https://github.com/openSUSE/zypper/pull/475#issuecomment-1382740585 for picture.

The download bar is already distinguishable from the preceding status line by the long streak of dashes/dots.


mlandres:
Thanks for the PR.

PR #482, opened 2023-08-27
Add missing include to fix build failure with new compiler in Debian

bluca:
```
 /<<PKGBUILDDIR>>/zypp/target/rpm/RpmHeader.cc: In function ‘int unameToUid(const char*, uid_t*)’:
 /<<PKGBUILDDIR>>/zypp/target/rpm/RpmHeader.cc:41:16: error: ‘strcmp’ was not declared in this scope
    41 |     } else if (strcmp(thisUname, "root") == 0) {
```

### DIFF
--- a/zypp/target/rpm/RpmHeader.cc
+++ b/zypp/target/rpm/RpmHeader.cc
@@ -11,6 +11,8 @@
 */
 #include "librpm.h"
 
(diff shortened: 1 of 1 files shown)


mlandres:
Thanks

PR #483, opened 2023-08-27
Prefer pkg-config to find gpgme

bluca:
gpgme-config has been deprecated and is not available in some distributions such as Debian, so prefer pkg-config and only use it as a fallback.


bzeller (approved):

bzeller (approved):
Thank you, all tests run, can be merged

PR #485, opened 2023-09-19
Use --runposttrans if rpm supports it

mlandres:
[bsc#1041742](https://bugzilla.suse.com/show_bug.cgi?id=1041742)
The initial commits pass the existing `RpmPostTransCollector` down to the `RpmDB` layer and provide a new `RpmDb::runposttrans` call for a delayed execution of  `%posttrans` and `%transfiletrigger(postun|in) `scripts.

Then `RpmPostTransCollector` is adapted to collect either `%posttrans` scripts or  `dump_posttrans:` lines, depending on what rpm supports. `RpmPostTransCollector::executeScripts` is adapted accordingly. Execution of `%posttrans` scripts by libzypp itself is intentionally not interruptible via callbacks because the same is true for `rpm --runposttrans` execuition.

Todo:
- [x] Handle rpm downgrade (collected dump_posttrans lines can no longer be executed by rpm)
- [x] Adapt discardScripts

### DIFF
--- a/zypp/target/RpmPostTransCollector.cc
+++ b/zypp/target/RpmPostTransCollector.cc
@@ -10,9 +10,13 @@
  */
 #include <iostream>
 #include <fstream>
+#include <optional>
 #include <zypp/base/LogTools.h>
 #include <zypp/base/NonCopyable.h>
 #include <zypp/base/Gettext.h>
+#include <zypp/base/Regex.h>
+#include <zypp/base/IOStream.h>
+#include <zypp/base/InputStream.h>
 #include <zypp/target/RpmPostTransCollector.h>
 
 #include <zypp/TmpPath.h>
@@ -20,7 +24,7 @@
 #include <zypp/HistoryLog.h>
 #include <zypp/ZYppCallbacks.h>
 #include <zypp/ExternalProgram.h>
-#include <zypp/target/rpm/RpmHeader.h>
+#include <zypp/target/rpm/RpmDb.h>
 #include <zypp/target/rpm/librpmDb.h>
 #include <zypp/ZConfig.h>
 #include <zypp/ZYppCallbacks.h>
@@ -35,7 +39,6 @@ namespace zypp
   ///////////////////////////////////////////////////////////////////
   namespace target
   {
-
     ///////////////////////////////////////////////////////////////////
     /// \class RpmPostTransCollector::Impl
     /// \brief RpmPostTransCollector implementation.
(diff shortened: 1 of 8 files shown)


zypp/UserData.h
```diff
@@ -49,6 +49,27 @@ namespace zypp
...
+        { userData_r.set( _key, std::move(val_r) ); }
+        ~ScopedKey() {
+          _userData.erase( _key );
+        }
+      private:
+        callback::UserData & _userData;
+        std::string _key;
+      };
+      ///////////////////////////////////////////////////////////////////
+
```
bzeller, line 72: We have zypp::OnScopeExit for that. Basically a AutoDispose<void> taking a lambda to clean up when leaving a scope. In the upcoming branch i even added some syntactical sugar around it

mlandres: Even better. I'll discard the class.

mlandres:
executeScripts is complete now (handles rpm downgrade case)

zypp/target/RpmPostTransCollector.cc
```diff
@@ -208,12 +335,59 @@ namespace zypp
...
+              ret = nullptr;
+          } else {
+            WAR << "Unexpectedly this is no package: " << rpmPackage_r << endl;
+          }
+          _headercache = std::make_pair( rpmPackage_r, ret );
+          return ret;
+        }
+
+        /** Retrieve "dump_posttrans: install" lines from \a dumpfile_r and pass n,v,r,a to the \a consumer_r. */
+        void recallFromDumpfile( const Pathname & dumpfile_r, std::function<void(std::string,std::string,std::string,std::string)> consume_r )
```
bzeller, line 370: For a impl class thats only used in the cc file anyway we could use a templated callback instead of std::function.
`std::is_invocable<void(string,string,string,string)>` can be used as constraint and to document the callable we need.

zypp/target/rpm/RpmDb.cc
```diff
@@ -1987,6 +2015,62 @@ void RpmDb::doRemovePackage( const std::string & name_r, RpmInstFlags flags, cal
...
+  MIL << "RpmDb::runposttrans(" << filename_r << ")" << endl;
+
+  RpmArgVec opts;
+  opts.push_back("-vv");  // want vverbose output to see scriptlet execution in the log
+  opts.push_back("--runposttrans");
+  opts.push_back(filename_r.c_str());
+  run_rpm (opts, ExternalProgram::Stderr_To_Stdout);
+
+  // Tailored to suit RpmPostTransCollector.
+  // It's a pitty, but we need all those verbose debug lines just
```
bzeller, line 2036:
```suggestion
  // It's a pity, but we need all those verbose debug lines just
```

bzeller (changes requested):
Some typos and a suggestion

zypp/target/RpmPostTransCollector.cc
```diff
@@ -208,12 +335,59 @@ namespace zypp
           return _ptrTmpdir->path();
         }
 
+        /** Return whether RpmHeader has a  %posttrans. */
+        bool headerHasPosttrans( rpm::RpmHeader::constPtr pkg_r ) const
+        {
+          bool ret = false;
+          if ( pkg_r ) {
+            std::string prog( pkg_r->tag_posttransprog() );
+            if ( not prog.empty() && prog != "<lua>" )  // by now leave lua to rpm
```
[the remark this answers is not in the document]
mlandres: Yes, that's the legacy behavior. We may review this later.

zypp/target/RpmPostTransCollector.cc
mlandres: Yes we could.  Though the method is only called once and this only if rpm was downgraded within the transaction to a version not supporting --runposttrans. So the runtime penalty is probably neglectable.

bzeller: Right

bzeller (approved):

PR #490, opened 2023-10-25
First pipelines

bzeller:
This patch adds the first async pipeline implementations.

### DIFF
--- a/zypp/KeyRing.h
+++ b/zypp/KeyRing.h
@@ -301,44 +301,6 @@ namespace zypp
     PublicKeyData trustedPublicKeyData( const std::string &id );
 
     /**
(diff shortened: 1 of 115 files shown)


zypp/KeyRing.h
```diff
@@ -26,6 +26,9 @@
 #include <zypp/PublicKey.h>
 #include <zypp/KeyRingContexts.h>
 
+#include <zypp-core/zyppng/async/AsyncOp>
+#include <zypp-core/zyppng/pipelines/Expected>
+
```
mlandres, line 31: /zyppng/ includes should not be exposed here (and are AFAICS not needed).

bzeller: Done

Issue #489, opened 2023-10-24
Gentoo build FAILED: tools/CMakeFiles/DownloadFiles.dir/DownloadFiles.cc.o


HougeLangley:
```
FAILED: tools/CMakeFiles/DownloadFiles.dir/DownloadFiles.cc.o 
/usr/bin/x86_64-pc-linux-gnu-g++ -DBOOST_PROGRAM_OPTIONS_DYN_LINK -DBOOST_PROGRAM_OPTIONS_NO_LIB -DBOOST_THREAD_DYN_LINK -DBOOST_THREAD_NO_LIB -DENABLE_ZCHUNK_COMPRESSION=1 -DHAVE_PIPE2 -DHAVE_RPM_VERIFY_TRANSACTION_STEP -DHAVE_UDEV -DPROTOBUF_USE_DLLS -DWITH_LIBPROXY_SUPPORT -DZYPP_RPM_BINARY=\"/usr/libexec/zypp/zypp-rpm\" -DZYPP_WORKER_PATH=\"/usr/libexec/zypp/workers\" -D_FILE_OFFSET_BITS=64 -I/var/tmp/portage/sys-libs/libzypp-17.31.22/work/libzypp-17.31.22 -I/var/tmp/portage/sys-libs/libzypp-17.31.22/work/libzypp-17.31.22_build -I/usr/include/rpm -I/usr/include/libxml2 -I/usr/include/sigc++-2.0 -I/usr/lib64/sigc++-2.0/include -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/libproxy -I/usr/lib64/libffi/include  -O2 -pipe -pthread -fno-strict-aliasing -fPIC -g -Wall -Wp,-D_GLIBCXX_ASSERTIONS -fvisibility-inlines-hidden -Woverloaded-virtual -Wnon-virtual-dtor -Werror=format-security -rdynamic -Wl,-as-needed  -std=c++17 -MD -MT tools/CMakeFiles/DownloadFiles.dir/DownloadFiles.cc.o -MF tools/CMakeFiles/DownloadFiles.dir/DownloadFiles.cc.o.d -o tools/CMakeFiles/DownloadFiles.dir/DownloadFiles.cc.o -c /var/tmp/portage/sys-libs/libzypp-17.31.22/work/libzypp-17.31.22/tools/DownloadFiles.cc
[...]

PR #495, opened 2023-11-13
tools/DownloadFiles: use standard zypp progress bar

jengelh:
I was just browsing commits and thought…

libzypp already contains a progress bar, no need to roll our own. (Demote PCBar to an adapter class.)

References: https://github.com/openSUSE/libzypp/issues/489


mlandres (commented):
ztui was just ripped out of zypper and is not yet intended for public use. The stuff in there is still quite zypper centric. Benjamin may judge whether he want's it in his tool.

bzeller (approved):
Yes, now that we have put ztui into libzypp we should use it for our internal tools as well. That was one of the reasons I moved it out of zypper into a standalone static lib

PR #497, opened 2023-11-17
Url varexpand issues

mlandres:
[bsc#1217192](https://bugzilla.suse.com/show_bug.cgi?id=1217192)


mlandres:
Introducing the `zypp-rawurl:` schema to transport unexpanded strings parsed from a .repo/.service file into the RepoInfo (which is able to actually replace the variables) has it's obstacles. We need to resolve all kinds of Url manipulation which were formerly performed on the unexpanded Url (based on the assumption they will be still valid after variable expansion).

Now we use the zypp-rawurl: query part to build a kind of todo list to be performed on the Url after variable expansion.

PR #501, opened 2023-12-08
Fix build issue with zchunk build flags (fixes #500)

### LINKED ISSUE #500: libzypp fails to build using cmake 3.28.0
From cmake 3.2.0's changelog:

```
+  * The "exec_program()" command, which has been deprecated since
+    CMake 3.0, has been removed by policy "CMP0153". Use the
+    "execute_process()" command instead.
```

cmake/modules/FindGpgme.cmake is making use of exec_program, which in turn makes build fail in some very 'fun' ways:

```
[   93s] 
[   93s]   /home/abuild/rpmbuild/BUILD/glibc-2.38/csu/../sysdeps/x86_64/start.S:103:(.text+0x2b):
[   93s]   undefined reference to `main'
[   93s] 
[   93s]   collect2: error: ld returned 1 exit status
[   93s] 
[   93s]   /bin/sh: line 1: -DOPENSSL_LOAD_CONF: command not found
[   93s] 
[   93s]   gmake[1]: *** [CMakeFiles/cmTC_18843.dir/build.make:78:
[   93s]   CMakeFiles/cmTC_18843.dir/printrpmver.c.o] Error 127
```

bzeller:
`PKG_CHECK_MODULES (ZCHUNK zck REQUIRED)` populates ZCHUNK_CFLAGS as a list and not a string. This resulted in semicolons ending up in CFLAGS which broke the build. Joining the variable before using it fixes the issue.


mlandres (approved):

DimStar77 (approved):
Short of knowing what exactly is the correct cmake way, I'd give this a go.

It seems quite strange for global cmake c flags to be modified like that (but I've found at least other projects doing similar hacks)

PR #509, opened 2024-01-08
Add a mechanism to signal to libzypp that a shutdown was requested

bzeller:


mlandres (approved):
OK. Just merge the `fixup!` commit.

PR #512, opened 2024-01-16
[L10n] Update th.po - more translations

bact:
- Update country names in Thai to use the official spelling from the Office of the Royal Society of Thailand
- Add some missing translations
- Fix few typos

### DIFF
--- a/po/th.po
+++ b/po/th.po
@@ -18,14 +18,14 @@ msgstr ""
 #: zypp/CountryCode.cc:50
 #, fuzzy
 msgid "Unknown country: "
-msgstr "ไม่ทราบประเภทของระเบียน: %1"
+msgstr "ไม่ทราบประเทศ: %1"
 
 #. Defined CountryCode constants
 #. Defined LanguageCode constants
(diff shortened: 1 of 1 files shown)


mlandres:
Thank you.

mlandres:
JFYI: Translations for [libzypp](https://l10n.opensuse.org/projects/libzypp/master/) and [zypper](https://l10n.opensuse.org/projects/zypper/master/) are _officially_ maintained in openSUSEWeblate. Changes applied there are automatically merged back into our git and your changes applied here should be uploaded to openSUSEWeblate. So I hope nothing gets lost.

bact:
Oh. Thank you. I should look at that instead.

(Just made PR #514 before I saw this)

Btw, a lot of translations are already avail in Unicode CLDR Project https://cldr.unicode.org/ - maybe openSUSEWeblate can utilize that as well (much better than machine translation).

mlandres:
@bact Frankly, I don't know how Weblate is working and whether it's possible to integrate third party services. But it's a great relief for us. 

(I merged #514)

PR #517, opened 2024-01-29
Add ProblemSolution::skipsPatchesOnly (for openSUSE/zypper#514)

mlandres:
Libzypp intentionally does not expose `SolutionAction`, so we offer it in `ProblemSolution`.

### DIFF
--- a/zypp/ProblemSolution.cc
+++ b/zypp/ProblemSolution.cc
@@ -23,6 +23,7 @@
  */
 
 #define ZYPP_USE_RESOLVER_INTERNALS
+#include <functional>
 
 #include <zypp/base/Gettext.h>
 #include <zypp/solver/detail/SolutionAction.h>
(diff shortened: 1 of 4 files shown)


zypp/ProblemSolution.cc
```diff
@@ -59,6 +60,17 @@ namespace zypp
     std::string		_details;
     SolutionActionList	_actions;
 
+    template <typename Predicate>
+    bool allActionsMatch( Predicate && predicate ) const
+    {
+      if ( _actions.empty() )
+        return false;
+      for ( const auto & aptr : _actions )
+        if ( not std::invoke( std::forward<Predicate>(predicate), aptr ) )
```
bzeller, line 69: Could perfect forwarding the predicate in a loop become a issue too?

mlandres: You mean std::invoke could somehow consume an _rvalue predicate_? I would not assume this.

bzeller (approved):

zypp/ProblemSolution.cc
bzeller: I was thinking about the possibility of a functor consuming itself, but that would also break if you pass that one by reference then...

PR #524, opened 2024-02-22
Prevent SINGLE_RPMTRANS from downloading packages if --dry-run.

mlandres:
Just the combination --dry-run --download-only will download packages in order to perform a file-conflict check.

@bzeller did we have a bug or issue number for this?

### DIFF
--- a/zypp/target/TargetImpl.cc
+++ b/zypp/target/TargetImpl.cc
@@ -1454,14 +1454,14 @@ namespace zypp
       bool singleTransMode = policy_r.singleTransModeEnabled();
 
       DBG << "commit log file is set to: " << HistoryLog::fname() << endl;
-      if ( ! policy_r.dryRun() || policy_r.downloadMode() == DownloadOnly || singleTransMode )
+      if ( ! policy_r.dryRun() || policy_r.downloadMode() == DownloadOnly )
       {
         // Prepare the package cache. Pass all items requiring download.
         CommitPackageCache packageCache;
         packageCache.setCommitList( steps.begin(), steps.end() );
 
         bool miss = false;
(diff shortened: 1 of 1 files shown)


zypp/target/TargetImpl.cc
```diff
@@ -1541,13 +1541,9 @@ namespace zypp
           {
             DBG << "dryRun/downloadOnly: Not installing/deleting anything." << endl;
             if ( explicitDryRun ) {
-              if ( policy_r.singleTransModeEnabled() ) {
-                // single trans mode does a test install via rpm
-                commitInSingleTransaction( policy_r, packageCache, result );
-              } else {
-                // if cache is preloaded, check for file conflicts
-                commitFindFileConflicts( policy_r, result );
-              }
```
mlandres, line 1550: Maybe we can even leave this in. If everything was downloaded, singletrans can check it as well.

zypp/target/TargetImpl.cc
```diff
@@ -1541,13 +1541,9 @@ namespace zypp
...
+              // fall back to libsolv's fileconflict check because signeltrans
```
bzeller, line 1544:
```suggestion
              // fall back to libsolv's fileconflict check because singletrans would require packages to be locally available
```

zypp/target/TargetImpl.cc
mlandres: Yes. It's the case where the download-all either succeeded or had nothing to do.

zypp/target/TargetImpl.cc
mlandres: Yes, but I reverted the change. It's the `miss == false` case, so all packages are available.

PR #529, opened 2024-03-06
Add default stripe minimum

Firstyear:
Due to ongoing issues with download speed, a minimum chunk/stripe download size is required to prevent conflict with TCP slowstart/congestion control.

I'm not 100% sure the patch is correct, the logic in this section is very convoluted and hard to follow. The idea is that if the calculated default blocksize is too small then we set it to a larger value. This way smaller files download in a single shot, but larger ones are chunked over multiple mirrors with a chunk size reasonable enough to actually get a TCP connection to it's full speed.

### DIFF
--- a/zypp/media/MediaMultiCurl.cc
+++ b/zypp/media/MediaMultiCurl.cc
@@ -297,7 +297,7 @@ constexpr auto MAXURLS       = 10;
 // but each range has enough time to increase it's congestion window to something reasonable.
 //
 // Initial value 4 MB;
-constexpr auto MIN_STRIPE_SIZE = 4194304;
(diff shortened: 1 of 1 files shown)


Firstyear:
> We already have a minimum chunk size but its currently 4K , as you can see in the `multifetchrequest::makeBlksize` function. We could just change the K to M there to get the same result

I couldn't find it sorry 🤣

But yes I'll adjust this. 4k is FAR too small for a minimum chunk size.

Firstyear:
@bzeller You'll need to squash-merge for this, I can't do a rebase because of #530 :(

mlandres:
Final tests for libzypp-17.32 are running. Probably next week.

Firstyear:
Sounds good, thanks!

PR #533, opened 2024-03-15
New vendor support option flag VendorSupportSuperseded

mlandres:
(jsc#OBS-301, jsc#PED-8014)


zypp/Package.cc
```diff
@@ -26,6 +27,61 @@ namespace zyppintern
...
+            case VendorSupportLevel2:
+              if ( kw == support_l3 )	{ ret.first = VendorSupportLevel3; break; }
+            case VendorSupportLevel3:
+              // VendorSupportSuperseded is special as the kw may also contain a '(PKGNAME)'.
+              // even more: multiple
+              if ( strv::hasPrefix( kw.asStringView(), support_superseded.asStringView() ) && kw.c_str()[kw.size()-1] == ')' ) {
+                ret.first = VendorSupportSuperseded;
+                if ( not ret.second )
+                  ret.second = std::set<std::string_view>();
+                ret.second->insert( std::string_view( kw.c_str()+support_superseded.size(), kw.size()-support_superseded.size()-1 ) );
```
bzeller, line 73: Not sure if I get this right, but once we encountered suuperseded once this case won't be entered again because ret.first is set to VendorSupportSuperseded ... so in the next for loop iteration the code would end up in the next case branch , so the Set will always only have one entry even if we encounter the superseded_by tag again

zypp/Package.cc
```diff
@@ -26,6 +27,61 @@ namespace zyppintern
...
+                return ret;
+              }
+            case VendorSupportSuperseded:
+              /* make gcc happy */ break;
```
bzeller, line 77: The code adding more string_views should go here

bzeller (changes requested):

zypp/Package.cc
mlandres: fixed

zypp/Package.cc
```diff
@@ -12,20 +12,81 @@
...
+              if ( kw == support_l2 )	{ ret.first = VendorSupportLevel2; break; }
+            case VendorSupportLevel2:
+              if ( kw == support_l3 )	{ ret.first = VendorSupportLevel3; break; }
+            case VendorSupportLevel3:
+            case VendorSupportSuperseded:
+              // VendorSupportSuperseded is special as the kw may also contain a '(PKGNAME)'.
+              // In fact even multiple (PKGNAMES) derived from multiple keywords.
+              if ( strv::hasPrefix( kw.asStringView(), support_superseded.asStringView() ) ) {
+                ret.first = VendorSupportSuperseded;
+                if ( kw.c_str()[support_superseded.size()] == '(' && kw.c_str()[kw.size()-1] == ')' ) {
```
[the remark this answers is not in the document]
mlandres: Yes, added one to the outer else as well.

bzeller (approved):

PR #535, opened 2024-03-18
repo2solv.sh is nowadays a symlink to repo2solv

dcermak:


mlandres:
Yes, but this branch serves the whole SLE15 family and might be in use to migrate SLE12->SLE15.  It would be a pity if we'd hit a system where the binary is not yet available. I'll need to check first what might be available in the customer systems first.

PR #547, opened 2024-05-06
Remove errnous version requires

dirkmueller:
the requires_eq is there to expand the version by itself, specifying it will not work


bzeller (approved):
Ok

PR #552, opened 2024-05-17
No Protobuf 

bzeller:


mlandres (approved):
But don't forget to remove `shared/tvm/zerocopystreams.*`  from git.

PR #554, opened 2024-06-10
Zypp Stack build

bzeller:
This patch adds the possibility to build libzypp as a subproject of a superproject build for the whole zypp stack.
Also uses a new approach to link tests statically to libzypp instead of dynamic so we do not need to export extra symbols for tests.


mlandres (changes requested):
Maybe a small README describing how to setup and use it.

mlandres (approved):

PR #557, opened 2024-06-30
Make sure the legacy APIConfig.h is installed too

Thaodan:


mlandres:
@Thaodan, sorry ;)
I saw the subject in my mailfolder and immediately fixed it. 
If I'd read the line up to it's end, I would have recognized that you in fact filed a PR and not just an issue.

PR #572, opened 2024-09-12
Require CMake 3.5

gruenich:
Newer versions of CMake warn that support for CMake < 3.5 is deprecated.


mlandres:
Thanks, I'll update zypper accordingly.

PR #575, opened 2024-09-13
Zyppng staging fix zypp task status

jreidinger:
With the fix, I am now able to generate proper rust bindings for all structs used in gir files.


bzeller (approved):

bzeller:
Thanks

PR #576, opened 2024-09-13
Reduce CMake warnings

gruenich:


cmake/modules/ZyppCommon.cmake
```diff
@@ -110,13 +110,7 @@ MACRO(GENERATE_PACKAGING PACKAGE VERSION)
...
   )
 
-  #SET(CPACK_PACKAGE_DESCRIPTION_SUMMARY "Novell's package management core engine.")
-  SET(CPACK_PACKAGE_VENDOR "Novell Inc.")
-  #SET(CPACK_PACKAGE_DESCRIPTION_FILE "${CMAKE_CURRENT_SOURCE_DIR}/ReadMe.txt")
-  #SET(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/Copyright.txt")
-  #SET(CPACK_PACKAGE_VERSION_MAJOR ${version_major})
-  #SET(CPACK_PACKAGE_VERSION_MINOR ${version_minor})
-  #SET(CPACK_PACKAGE_VERSION_PATCH ${version_patch})
+  SET(CPACK_PACKAGE_VENDOR "SUSE LLE")
```
mlandres, line 113:
```suggestion
  SET(CPACK_PACKAGE_VENDOR "SUSE LLC")
```

mlandres (changes requested):

cmake/modules/ZyppCommon.cmake
gruenich: Thanks for pointing this out, I fixed it without an additional commit.

mlandres (approved):
Danke!

PR #577, opened 2024-09-15
Fix a couple of Cppcheck warnings

gruenich:


tools/zypp-rpm/BinHeader.cc
```diff
@@ -90,21 +90,21 @@ unsigned BinHeader::intList::set( void * val_r, unsigned cnt_r, rpmTagType type_
     {
 #if RPM_CHAR_TYPE != RPM_INT8_TYPE
       case RPM_CHAR_TYPE:
-        std::vector<long>( (char*)val_r, ((char*)val_r)+cnt_r ).swap( _data );
+        std::vector<long>( static_cast<char*)val_r, static_cast<char*>(val_r)+cnt_r ).swap( _data );
```
mlandres, line 93:
```suggestion
        std::vector<long>( static_cast<char*>(val_r), static_cast<char*>(val_r)+cnt_r ).swap( _data );
```

mlandres (changes requested):
Apparently we have no distro with RPM5 in outr checks (if there is any at all)

PR #581, opened 2024-09-30
Add predownload plugin

andrii-suse:
Purpose:

  *  get faster downloads utilizing concurrent downloads with sypplite utility
  *  let others experiment with custom implementation of download (e.g. to deal with their specific limitations)

Implementation algorithm:

  * check if any plugin present;
  * create list of Transaction Steps which require download;
  *  for each repository involved in future downloads: send to plugin a string where first word is baseurl and the tail is list of files required for download from that repo. (command PREDOWNLOAD_FROM_REPO);
   * if any of plugins is still present: pick the first one and ask him to do actual download (command PREDOWNLOAD_START);
   * check plugin for progress every second (command PLUGIN_PROGRESS);
   * All packages are expected to be in the cache with a suffix ".unverified" now;
   * The fetcher code now is aware of ".unverified" suffix and tries to look up and verify such files;

Implementation details:

   * a new method progress() was added to PluginScript, which defines how plugins will report progress;
   * the current implementation of the plugin [sypplite.sh](https://github.com/andrii-suse/sypper/blob/master/predownload/sypplite.sh) will create request files for sypplite utility in /var/cache/zypp/packages/ .

Testing with prebuilt libzypp (Tumbleweed only so far):
```
zypper ar -f http://download.opensuse.org/repositories/home:andriinikitin:sypper/openSUSE_Tumbleweed home:andriinikitin:sypper
zypper --gpg-auto-import-keys  ref
zypper -n in --from home:andriinikitin:sypper sypplite libzypp
cp /usr/share/sypper/predownload/sypplite.sh /usr/lib/zypp/plugins/predownload/
```

Current experimented results for performance gain on 200 - 3000 packages is 40% - 200%

### DIFF
--- a/zypp/Fetcher.cc
+++ b/zypp/Fetcher.cc
@@ -196,7 +196,7 @@ namespace zypp
        * the cache (matching checksum is mandatory). Returns the
        * location of the cached file or an empty \ref Pathname.
        */
-      Pathname locateInCache( const OnMediaLocation & resource_r, const Pathname & destDir_r );
+      Pathname locateInCache( const OnMediaLocation & resource_r, const Pathname & destDir_r, const std::string & suffix_in_cache = "" );
       /**
        * Validates the provided file against its checkers.
        * \throws Exception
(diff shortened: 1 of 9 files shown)


lkocman:
40% is super impressive!

lkocman:
I'm not sure what's the teams capacity, but using a plugin and over the time merge functionality into the base libs could be a good short term solution. This woud allow us to aim at 16.0 Beta and similar.

andrii-suse:
closing for now in favor of https://github.com/openSUSE/libzypp/pull/623

PR #583, opened 2024-10-08
Path for update messages and scripts shouldn't be user configurable

bzeller:


zypp/ZConfig.h
```diff
@@ -496,19 +496,17 @@ namespace zypp
       /**
        * Path where the update items are kept (/var/adm)
        */
-      Pathname update_dataPath() const;
+      static Pathname update_dataPath();
```
mlandres, line 499:
```suggestion
#if LEGACY(1735)
      Pathname update_dataPath() const;
#else
      static Pathname update_dataPath();
#endif
```

mlandres: And accordingly for the other two and in the .cc file.
So we don't need to break the API immediately.

mlandres (changes requested):

zypp/ZConfig.h
bzeller: Done

PR #584, opened 2024-10-09
PluginFrame: Send unescaped colons in header values (bsc#1231043)

mlandres:
According to the STOMP protocol it would be correct to escape a colon in a header-value, but it breaks plugin receivers which do not expect this. The first colon separates header-name from header-value, so escaping in the header-value is not needed anyway.

Escaping in the header-value affects especially the urlresolver plugins. The input URL is passed in a header, but sent back as raw data in the frames body. If the plugin receiver does not correctly unescape the URL we may get back a "https\c//" which is not usable.

### DIFF
--- a/zypp-core/rpc/PluginFrame.cc
+++ b/zypp-core/rpc/PluginFrame.cc
@@ -74,7 +74,7 @@ namespace zypp
       void setBody( ByteArray && body_r )
       { _body = std::move(body_r); }
 
-      static std::string escapeHeader( std::string_view val ) {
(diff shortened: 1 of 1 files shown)


bzeller (approved):
FTR I still would've preferred a explicit compat mode in PluginFrame over sending  invalid header according to spec implicitely by default. Not gonna block on that though.

mlandres:
> FTR I still would've preferred a explicit compat mode in PluginFrame over sending invalid header according to spec 

Well, it's still valid STOMP 1.0.  And even 1.1 and 1.2 do not escape it in e.g. CONNECT and CONNECTED frames.

PR #585, opened 2024-10-10
Use std::shared_ptr and std::weak_ptr instead of their ancestors from Boost

gruenich:
C++ 11 introduced `std::shared_ptr` and `std::weak_ptr`. Those pointer types were introduced in Boost and included in the C++ standard.

This pull request is a first draft to see if the project is interested in such type of changes and to decide on the direction to go.

Known issues:
[ ] No consideration of deprecation for other parts or project depending on `zypp:*_ptr`.
[ ] There might be more things in `PtrTypes.h` that can be clean up.
[ ] Tests are not adapted, nor compiled.
[ ] Software depending on libzypp might need adjustments, too.

### DIFF
--- a/tests/lib/FtpServer.cc
+++ b/tests/lib/FtpServer.cc
@@ -314,7 +314,7 @@ public:
 
     filesystem::TmpDir _workingDir;
     zypp::Pathname _docroot;
-    zypp::shared_ptr<std::thread> _thrd;
+    std::shared_ptr<std::thread> _thrd;
 
     unsigned int _port;
     int _wakeupPipe[2];
--- a/tests/lib/WebServer.cc
+++ b/tests/lib/WebServer.cc
@@ -24,6 +24,7 @@
 
 #include <thread>
 #include <atomic>
+#include <memory>
 #include <mutex>
 #include <condition_variable>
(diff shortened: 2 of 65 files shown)


bzeller:
Since we are currently refactoring big parts of libzypp a big PR like that must wait... we moved a lot of code around in staging. So big patches like this must need to wait until we have the refactoring finished and in master.

gruenich:
Thanks for pointing out the current state. Once your re-org is done, I will create another merge request.

mlandres:
Please don't touch the smart pointer types. It's an unnecessary API breakage.

And ideally (if there wasn't some code meanwhile using explicit `boot::*:ptr`), the switch could mainly be done by simply changing the using directives in [zypp-core/base/PtrTypes.h](https://github.com/openSUSE/libzypp/blob/master/zypp-core/base/PtrTypes.h). It's been the intent of the file to define what a `zypp::*_ptr` actually is.

gruenich:
You are right. I replaced all remaining uses of `boost::*_ptr` in #605. After we cleaned up some consumers (zypper, Myrlyn probably Yast Software), we can consider a proper transition (like a fallback with a CMake option).

PR #586, opened 2024-10-11
Replace some Boost code by C++11

gruenich:
* `<functional>`
* static_assert
* noexcept


mlandres (commented):
We should IMO put it on hold as it touches locations which are under refactoring or going to be deleted anyway (like ParserProgress)

gruenich:
@mlandres Is now a better time to consider merging?

gruenich:
Rebased, should be good to be merged.

PR #590, opened 2024-11-22
cmake: check location of fcgi header and adjust include accordingly

bluca:
On Debian and derivatives the fcgi headers are not stored in a fastcgi/ subdirectory:

https://packages.debian.org/trixie/amd64/libfcgi-dev/filelist


mlandres:
Thanks!

PR #592, opened 2024-12-03
Fix small typo

grisu48:
Replace `permision` with `permission` in the language files.

### DIFF
--- a/po/af.po
+++ b/po/af.po
@@ -4150,7 +4150,7 @@ msgid "applydeltarpm failed."
 msgstr "applydeltarpm het misluk."
 
 #: zypp/repo/RepoException.cc:73
-msgid "No permision to write repository cache."
+msgid "No permission to write repository cache."
 msgstr ""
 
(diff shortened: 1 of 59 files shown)


mlandres:
Thanks, but that's not the correct way to fix it. Those .po files are at the end of a translation chain. 
```
#: zypp/repo/RepoException.cc:73
msgid "No permision to write repository cache."
```
The comment hints to the source code file `zypp/repo/RepoException.cc` which contains the line with the typo:
```
    DEF_CTORS( RepoNoPermissionException,   _("No permision to write repository cache.") );
```
The `_(...)` macro in our code tags the texts which should be translated. Those texts are extracted from the source code into the `po/zypp.pot` file. The `.pot` file is the source for the translations. The `.po` files are the result of translating the .pot file texts into the individual languages. At runtime the text in the source code is used to lookup the translation in the loaded .po file.

I'll fix the typo in the source code and recreate the .pot and .po files.

PR #594, opened 2024-12-08
Url query part: `=` is a safe char in value

mlandres:
([bsc#1234304](https://bugzilla.suse.com/show_bug.cgi?id=1234304))


mlandres:
Will add another commit to prevent decoding-reencoding sequences in `delQueryParam`. This is to preserve the original querypart as long as possible (in case it intentionally uses a mix of  clear and encoded safe chars thought his would be kind of broken kind of broken by design IMO _?token=exp`=`1370627409~acl`=`?*`%3D`*~hmac`=`6a10b3f..._)

PR #597, opened 2024-12-17
Set default value for config option 'download_use_deltarpm' to false

rhabacker:
Nowadays, network bandwidth is no longer so much of an issue, so changing the default setting can save memory and CPU time when downloading.

See #581


mlandres:
JFYI: see #620 Disable deltarpm by default
Leap/SLES 15.7 will have deltas disabled by default in zypp (but AFAIK still offers them)
Leap/SLES 16 AFIAK not even offers them anymore

PR #599, opened 2025-01-05
Fix four sign-compare warnings

gruenich:
Found by GCC 15.

### DIFF
--- a/zypp-curl/parser/mediablocklist.cc
+++ b/zypp-curl/parser/mediablocklist.cc
@@ -492,7 +492,7 @@ void MediaBlockList::reuseBlocks(FILE *wfp, const std::string& filename)
       // build the hashtable
       uint rsumHashMask = 0;
       {
-        int i = 16;
+        uint i = 16;
 
         /* Try hash size of 2^i; step down the value of i until we find a good size */
         while ((2 << (i - 1)) > nblks && i > 4)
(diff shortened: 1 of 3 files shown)


mlandres:
Sorry, but they can't simply change the signedness of a variable to make a compiler warning go away ;)

`int i; if ( i >= 0 ...` is meaningful, while
`uint i; if ( i >= 0 ...` isn't.

The compiler might be happy, but the code would be wrong. Changing the signedness of a variable is not a trivial change. 
They need to understand why the variable is signed (or unsigned) and check _all_ locations where the variable is used.

Whether one likes it or not, using `int` for naturally `unsigned` values in order to let `-1` indicate some kind of _error_ or _null value_ is a concept you'll find quite often. Personally I'd prefer using  `unsigned`  and an explicit `null_value = unsigned(-1)`  constant. Both is possible but leads to different code.

mlandres:
Sorry, no need to close it. But it needs to be fixed.

mlandres (changes requested):

gruenich:
Thank you for spotting my wrongly changed type! I replaced it with a `static_cast`. I also adjusted the other type as you suggested.

zypp-curl/parser/mediablocklist.cc
```diff
@@ -492,7 +492,7 @@ void MediaBlockList::reuseBlocks(FILE *wfp, const std::string& filename)
       // build the hashtable
       uint rsumHashMask = 0;
       {
-        int i = 16;
+        uint i = 16;
 
         /* Try hash size of 2^i; step down the value of i until we find a good size */
         while ((2 << (i - 1)) > nblks && i > 4)
```
mlandres, line 498: A bit late after merge, but did you compile it? 
The expression is still ` int` because the `2` is an 'int'.  `2U << ...` would have done the trick.

gruenich: I compiled it and I think the sign warning went away, but I am no longer sure. Thanks for your change fixing the two shift cases!

PR #607, opened 2025-01-25
Reduce number of Clang warnings

gruenich:
Some are only part of `-Wall` or `-Wpedantic`.

### DIFF
--- a/zypp-logic/zypp/ng/repo/workflows/repodownloaderwf.cc
+++ b/zypp-logic/zypp/ng/repo/workflows/repodownloaderwf.cc
@@ -108,7 +108,7 @@ namespace zyppng {
                   return makeReadyResult(expected<void>::success());
                 };
               })
-            | [this,masterres=std::move(masterres)]( expected<void> ) {
+            | [masterres=std::move(masterres)]( expected<void> ) {
               return make_expected_success( std::move(masterres) );
             };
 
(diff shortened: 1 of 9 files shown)


zypp/media/MediaPriority.cc
mlandres, line 1: No. Please leave my switch untouched.

mlandres (changes requested):

zypp/media/MediaPriority.cc
gruenich: 1. Is it okay that I replace the macro `RETURN_IF` by direct `if`s with `std::string::compare` and `return`s?
2. Should I replace the inside the macro `RETURN_IF` the `::strcmp` by `std::string::compare`?
3. Should I leave the whole `MediaPriority.cc` as it is, even if subjectively it is more difficult to maintain that it has to be?

gruenich: I went with 1. and removed the macro `RETURN_IF` but preserved the switch.

gruenich:
@mlandres Do you mind having another look?

zypp/media/MediaPriority.cc
mlandres: Sorry, 3rd. please. Leave the switch as it is.

When adding `override` please remove the `virtual`. I think we don't keep both anywhere.

gruenich: Thanks for the feedback! I updated this pull request accordingly.

gruenich:
Rebased, ready to be merged.

PR #618, opened 2025-02-13
Experimental Mediacurl backend

bzeller:


mlandres:
While you're at it: In section [zypp-envars-logging Variables related to logging[(https://github.com/openSUSE/libzypp/blob/master/doc/autoinclude/EnvironmentVariables.doc#L33) the entries for `ZYPP_MEDIANETWORK` and `ZYPP_MULTICURL` should be removed. They are in [zypp-envars-mediabackend Selecting the mediabackend to use](https://github.com/openSUSE/libzypp/blob/master/doc/autoinclude/EnvironmentVariables.doc#L41) already. There we should also mention `ZYPP_CURL2`.

mlandres (approved):

PR #621, opened 2025-02-21
Filesystem usrmerge must not be done in singletrans mode

mlandres:
[bsc#1236481](https://bugzilla.suse.com/show_bug.cgi?id=1236481), [bsc#1189788](https://bugzilla.suse.com/show_bug.cgi?id=1189788)

Commit will amend the backend in case the transaction would perform a filesystem usrmerge.


mlandres:
@bzeller In case of doubt I'll release it as a quick fix. The usrmerge checks should IMO be done in the Target, but this requires some extra sanitizeCommitPolicy call if zypper wants to report enforced changes before the commit prompt.

PR #628, opened 2025-03-03
Curl2 preload improvements

bzeller:


mlandres (approved):

zypp/RepoInfo.cc
```diff
@@ -681,6 +758,13 @@ namespace zypp
   bool RepoInfo::baseUrlSet() const
   { return _pimpl->baseurl2dump(); }
 
+  Url RepoInfo::url() const
+  {
+    if ( _pimpl->effectiveBaseUrls().empty() )
+      return Url();
+    return _pimpl->effectiveBaseUrls().front();
+  }
```
mlandres, line 766: I'd suggest to revert this. First of all: `rawUrl()` and `url()` need to be in sync. `url()` should be _the url_ applications  associate with the repo. This shoudl be the baseurl.
What we could do, in case baseurl is empty and mirror/metalink url is not, using this as fallback. But I would not use one of the expanded mirror list urls. The value should stay fix, no matter what mirrors are currently provided for it.

PR #629, opened 2025-03-14
Fix lost double slash when appending to an absolute FTP url

mlandres:
[bsc#1238315](https://bugzilla.suse.com/show_bug.cgi?id=1238315)

Ftp actually differs between absolute and relative URL paths. Absolute path names begin with a double slash encoded as '/%2F'. This must be preserved when manipulating the path.


bzeller (approved):
OK

PR #634, opened 2025-04-16
Enable service to set more <repo> related attributes

mlandres:


zypp/parser/RepoindexFileReader.cc
```diff
@@ -180,6 +180,11 @@ namespace zypp
       // xpath: /repoindex/data (+)
       if ( reader_r->name() == "repo" )
       {
+        // TODO: Ideally the repo values here are interpreted the same way as
+        // corresponding ones in the RepoFileReader. Check whether we can introduce
+        // a RepoInfo ctor from a string dict. Using it here and in RepoFileReader.
+        // For now we try to parse the same was as in RepoFileReader.
```
bzeller, line 186:
```suggestion
        // For now we try to parse the same way as in RepoFileReader.
```

zypp/parser/RepoindexFileReader.cc
```diff
@@ -245,9 +249,34 @@ namespace zypp
...
+        if ( getAttrValue( "gpgkey", reader_r, attrValue ) )
+          info.setGpgKeyUrl( Url(attrValue) );
+
+        // optional mirrorlist
+        if ( getAttrValue( "mirrorlist", reader_r, attrValue ) )
+          info.setMirrorListUrl( Url(attrValue) );
+
+        // optional metalink
+        if ( getAttrValue( "metalink", reader_r, attrValue ) )
+          info.setMetalinkUrl( Url(attrValue) );
```
bzeller, line 274: Should we support only one OR the other? IIRC RepoInfo would not support both... alternatively we could let the new  `effectiveBaseUrls` function in my PR just unite all the lists into one

bzeller (changes requested):

zypp/parser/RepoindexFileReader.cc
mlandres: I followed the `ReopoInfoReader`. It also accepts both and sets both (first mirrorlist then metalink). `RepoInfo` internally uses just the last one set, so metalink superseeds mirrorlist. It's not different from specifying a baseurl in together with a metalink url. It's ok, the metalinkurl superseeds the baseurl.

zypp/parser/RepoindexFileReader.cc
mlandres: fixed

PR #643, opened 2025-06-04
Append a trailing slash to baseUrl

dirkmueller:
The previous code appendPathName() wasn't doing anything because that function is ignoring empty or "/" as parameter. so do it opencoded instead.

### DIFF
--- a/zypp/RepoInfo.cc
+++ b/zypp/RepoInfo.cc
@@ -186,7 +186,9 @@ namespace zypp
            && repo::RepoMirrorList::urlSupportsMirrorLink( *_baseUrls.transformedBegin() ) ) {
 
         mlurl = *_baseUrls.transformedBegin ();
-        mlurl.appendPathName("/");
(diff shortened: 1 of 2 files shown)


zypp/RepoInfo.cc
```diff
@@ -186,7 +186,9 @@ namespace zypp
            && repo::RepoMirrorList::urlSupportsMirrorLink( *_baseUrls.transformedBegin() ) ) {
 
         mlurl = *_baseUrls.transformedBegin ();
-        mlurl.appendPathName("/");
+        if (!mlurl.getPathName().empty() && mlurl.getPathName() != "/" && !zypp::str::endsWith(mlurl.getPathName(), "/") ) {
```
josegomezr, line 189: Silly comment prolly but the expression condition can also be catched by the last expression in the if (endswith slash returns true if the value if a lonely slash).

```suggestion
        if (!mlurl.getPathName().empty() && !zypp::str::endsWith(mlurl.getPathName(), "/") ) {
```

mlandres:
I'll add a dedicated method for this to the Url class.

PR #651, opened 2025-06-18
Fix evaluation of libproxy results (bsc#1244710)

mlandres:


mlandres:
@bzeller I changed the code to behave like proxyinfosysconfig. We return a defined proxy with the same schema and not just `http:`. the comment that _"cURL can only handle HTTP proxies"_ is IMO a legacy one. `CURLOPT_PROXY(3)` says https was added in 7.52.0.

PR #659, opened 2025-07-11
MediaCurl2: Download zck without pre known headersize

bzeller:
This patch teaches MediaCurl2 to download files without previously knowing the zchunk header size, a situation we have usually for metadata files. It requires one extra request to download the lead to know the actual header size.


mlschroe:
I don't think you should do such a small request, as requests are expensive. You should instead get a "reasonable" chunk and hope that it is big enough to contain the complete header. The idea is that doing a second request should be rare.

mlschroe:
Requests are expensive even if the connection is reused.

PR #663, opened 2025-07-17
Allow explicit request to probe an added repo's URL  

mlandres:
[bsc#1246466] (https://bugzilla.suse.com/show_bug.cgi?id=1246466)


bzeller (approved):
Maybe consider to switch from std::optional<bool> to TriBool to better express the intention, otherwise code is fine.

mlandres:
Changed it to TriBool.

PR #694, opened 2026-01-22
Fall back to a writable location when precaching packages without root

danidoni:
Until now, when using the --installroot option and the preload mechanism, we get "could not create target file" errors during the preloading phase because zypper tries to write into the general packages cache directory (`/var/cache/zypp/packages/repo-oss/.preload`).

With this change, we detect the lack of permissions and fallback to a writable temporary cache directory, somewhere in
`/var/tmp/zypp.something/var/cache/zypp/packages`.

bsc#1247948

### DIFF
--- a/zypp-logic/zypp/repo/RepoProvideFile.cc
+++ b/zypp-logic/zypp/repo/RepoProvideFile.cc
@@ -126,6 +126,20 @@ namespace zypp
       return access.provideFile(std::move(repo_r), loc_r, policy_r );
     }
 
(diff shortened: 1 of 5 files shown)


zypp-logic/zypp/repo/RepoProvideFile.cc
bzeller: Since this is using packagesPath() it will not add the system level cache in case we are a normal user. We should also find packages that are possibly cached in the system level cache if we are a normal user

bzeller (approved):
Looks good from my side

PR #699, opened 2026-02-11
Support multiple MirroredOrigin authorities

danidoni:
Support multiple MirroredOrigin authorities.

When multiple baseurls are configured into a repo only the first one could be an authority, the rest were treated as mirrors. Because falling back to the mirrors is not allowed when downloading the master index (the repomd.xml file for the configured repo), if the authority was unreachable, commands like zypper ref would fail. 

With this change, if the first authority is unreachable, we can fall back to any other authority, if defined.

bsc#1253193

### DIFF
--- a/zypp-logic/zypp-media/ng/lazymediahandle.h
+++ b/zypp-logic/zypp-media/ng/lazymediahandle.h
@@ -46,7 +46,7 @@ namespace zyppng
      * Returns the first mirror used for this lazy handle,
      * otherwise returns a empty url.
      */
-    const zypp::Url baseUrl() const {
+    zypp::Url baseUrl() const {
       if ( !_sharedData->_mediaHandle ) {
         return _sharedData->_origin.authority().url();
       } else {


zypp-logic/zypp-core/MirroredOrigin.cc
```diff
@@ -193,19 +205,56 @@ namespace zypp {
 
   bool MirroredOrigin::isValid() const
   {
-    return _pimpl->_authority.isValid();
+    return _pimpl->authoritiesAreValid();
+  }
+
+  bool MirroredOrigin::addAuthority(OriginEndpoint newAuthority)
+  {
+    if ( _pimpl->authoritiesAreValid() ) {
```
bzeller: Add a check at the beginning if the newAuthority is valid , otherwise we do not need to check further: 
if ( !newAuthority.isValid() ) { /* write error to log and return */ }

zypp-logic/zypp-core/MirroredOrigin.cc
```diff
@@ -193,19 +205,56 @@ namespace zypp {
...
+           && ( authScheme != newAuthority.scheme () )
+      ) {
+        MIL << "Ignoring authority " << newAuthority << " scheme is not compatible to current authority URL ( " << newAuthority.scheme() << " vs " << authScheme << ")" << std::endl;
+        return false;
+      }
+    }
+
+    if (std::find(_pimpl->_authorities.begin(), _pimpl->_authorities.end(), newAuthority) != _pimpl->_authorities.end()) {
+      MIL << "Ignoring authority " << newAuthority << " already present in authorities" << std::endl;
+      return true;
```
bzeller: Should we here also check if the URL is already in mirrors? And if yes should we drop it from there and use it as authority instead?

PR #719, opened 2026-03-26
Fix different schemas for authorities and mirrors

danidoni:
When an authority is a downloading one and its mirrors is not (or the other way around) it's scheme is considered incompatible, so we clean its mirrors to not have mixed schemes.

### DIFF
--- a/zypp-logic/zypp-core/MirroredOrigin.cc
+++ b/zypp-logic/zypp-core/MirroredOrigin.cc
@@ -176,7 +176,6 @@ namespace zypp {
     }
 
     const auto &newScheme = newAuthority.scheme();
-    bool newAuthIsDl = newAuthority.url().schemeIsDownloading();
 
     _pimpl->_authorities.clear();
     _pimpl->_authorities.push_back( std::move(newAuthority) );
@@ -186,9 +185,9 @@ namespace zypp {
 
     // house keeeping, we want only compatible mirrors
     for ( auto i = _pimpl->_origins.begin (); i != _pimpl->_origins.end(); ) {
-      if (    ( newAuthIsDl && !i->schemeIsDownloading() ) // drop mirror if its not downloading but authority is
-           && ( i->scheme () != newScheme )                // otherwise drop if scheme is not identical
-      ) {
+      // drop mirror if it's scheme conflicts with the authorities scheme
+      // otherwise drop if scheme is not identical
+      if ( !areAuthoritiesCompatible(*i) && ( i->scheme () != newScheme ) ) {
         MIL << "Dropping mirror " << *i << " scheme is not compatible to new authority URL ( " << i->scheme() << " vs " << newScheme << ")" << std::endl;
         i = _pimpl->_origins.erase(i);
       } else {
@@ -200,6 +199,23 @@ namespace zypp {
   const std::vector<OriginEndpoint> &MirroredOrigin::authorities() const
   {
     return _pimpl->_authorities;
+  }
+
+  bool MirroredOrigin::isAuthorityCompatible(const OriginEndpoint& authority, const OriginEndpoint& mirror)
+  {
+    return (authority.schemeIsDownloading() && mirror.schemeIsDownloading())
+        || (!authority.schemeIsDownloading() && !mirror.schemeIsDownloading());
+  }
+
+  bool MirroredOrigin::areAuthoritiesCompatible(const OriginEndpoint& mirror)
+  {
+    for (auto authority: _pimpl->_authorities)
+    {
+      if (!isAuthorityCompatible(authority, mirror))
+        return false;
+    }
+
+    return true;
   }
 
   const OriginEndpoint &MirroredOrigin::authority() const
@@ -258,11 +274,9 @@ namespace zypp {
 
     if ( _pimpl->authoritiesAreValid() ) {
       const auto &authScheme = _pimpl->_authorities[0].scheme();
-      bool authIsDl = _pimpl->_authorities[0].schemeIsDownloading();
 
-      if ( ( authIsDl && !newMirror.schemeIsDownloading () )
-           && ( authScheme != newMirror.scheme () )
-      ) {
+      // As all the authorities are either downloading or non-downloading, if the first is compatible, the rest are as well
+      if ( !isAuthorityCompatible(authority(), newMirror) && ( authScheme != newMirror.scheme () ) ) {
         MIL << "Ignoring mirror " << newMirror << " scheme is not compatible to authority URL ( " << newMirror.scheme() << " vs " << authScheme << ")" << std::endl;
         return false;
       }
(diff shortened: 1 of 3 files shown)


zypp-logic/zypp-core/MirroredOrigin.cc
```diff
@@ -189,6 +188,12 @@ namespace zypp {
     }
   }
 
+  bool MirroredOrigin::isCompatible(const OriginEndpoint& authority, const OriginEndpoint& mirror)
+  {
+    return (authority.schemeIsDownloading() && mirror.schemeIsDownloading())
+        || (!authority.schemeIsDownloading() && !mirror.schemeIsDownloading());
```
mlandres, line 207:
```suggestion
    return authority.schemeIsDownloading() == mirror.schemeIsDownloading();
```
schemeIsDownloading is a boolean.

mlandres: Additionally the method could be inline and should be `const`, or even better `static`.

zypp-logic/zypp-core/MirroredOrigin.cc
```diff
@@ -202,6 +201,23 @@ namespace zypp {
...
+  bool MirroredOrigin::areAuthoritiesCompatible(const OriginEndpoint& mirror)
+  {
+    for (auto authority: _pimpl->_authorities)
+    {
+      if (!isAuthorityCompatible(authority, mirror))
+        return false;
+    }
+
+    return true;
+  }
```
mlandres, line 219: Method should be `const` and the for loop should use a `const&`.

zypp-logic/zypp-core/MirroredOrigin.cc
```diff
@@ -186,9 +185,9 @@ namespace zypp {
 
     // house keeeping, we want only compatible mirrors
     for ( auto i = _pimpl->_origins.begin (); i != _pimpl->_origins.end(); ) {
-      if (    ( newAuthIsDl && !i->schemeIsDownloading() ) // drop mirror if its not downloading but authority is
-           && ( i->scheme () != newScheme )                // otherwise drop if scheme is not identical
-      ) {
+      // drop mirror if it's scheme conflicts with the authorities scheme
+      // otherwise drop if scheme is not identical
+      if ( !areAuthoritiesCompatible(*i) && ( i->scheme () != newScheme ) ) {
```
mlandres, line 190: **suggestion**
```
      if ( i->scheme() != newScheme && !areAuthoritiesCompatible(*i) ) {
```
Easiest check first. No need to loop if schemata are equal.
OTOH - due to L180,181 - you know there's just one authority, so `isAuthorityCompatible` would do it as well.

zypp-logic/zypp-core/MirroredOrigin.cc
```diff
@@ -258,11 +274,9 @@ namespace zypp {
 
     if ( _pimpl->authoritiesAreValid() ) {
       const auto &authScheme = _pimpl->_authorities[0].scheme();
-      bool authIsDl = _pimpl->_authorities[0].schemeIsDownloading();
 
-      if ( ( authIsDl && !newMirror.schemeIsDownloading () )
-           && ( authScheme != newMirror.scheme () )
-      ) {
+      // As all the authorities are either downloading or non-downloading, if the first is compatible, the rest are as well
+      if ( !isAuthorityCompatible(authority(), newMirror) && ( authScheme != newMirror.scheme () ) ) {
```
mlandres, line 279: As above, easiest test first. No compat check for equal schemata.

mlandres (changes requested):

zypp-logic/zypp-core/MirroredOrigin.cc
danidoni: Fixed

zypp-logic/zypp-core/MirroredOrigin.cc
danidoni: Fixed

zypp-logic/zypp-core/MirroredOrigin.cc
danidoni: Fixed

zypp-logic/zypp-core/MirroredOrigin.cc
danidoni: Fixed

mlandres:
@danidoni You did not compile or your submission is incomplete. 

> [   67s] /home/abuild/rpmbuild/BUILD/libzypp-18.0.0.git.1777360702.ed8c808/zypp-logic/zypp-core/MirroredOrigin.cc:204:8: error: prototype for 'bool zypp::MirroredOrigin::areAuthoritiesCompatible(const zypp::OriginEndpoint&) const' does not match any in class 'zypp::MirroredOrigin'
[   67s]    bool MirroredOrigin::areAuthoritiesCompatible(const OriginEndpoint& mirror) const
[   67s]         ^~~~~~~~~~~~~~

danidoni:
> @danidoni You did not compile or your submission is incomplete.
> 
> > [   67s] /home/abuild/rpmbuild/BUILD/libzypp-18.0.0.git.1777360702.ed8c808/zypp-logic/zypp-core/MirroredOrigin.cc:204:8: error: prototype for 'bool zypp::MirroredOrigin::areAuthoritiesCompatible(const zypp::OriginEndpoint&) const' does not match any in class 'zypp::MirroredOrigin'
> > [   67s]    bool MirroredOrigin::areAuthoritiesCompatible(const OriginEndpoint& mirror) const
> > [   67s]         ^~~~~~~~~~~~~~

I forgot to add a last minute change.

zypp-logic/zypp-core/MirroredOrigin.cc
```diff
@@ -202,6 +201,23 @@ namespace zypp {
     return _pimpl->_authorities;
   }
 
+  bool MirroredOrigin::isAuthorityCompatible(const OriginEndpoint& authority, const OriginEndpoint& mirror)
+  {
+    return (authority.schemeIsDownloading() && mirror.schemeIsDownloading())
+        || (!authority.schemeIsDownloading() && !mirror.schemeIsDownloading());
```
bzeller: A Authority is only compatible if both are downloading. In all other cases they have to be equal e.g.:

HTTP vs FTP -> compatible ( both are downloading )
HTTP vs HTTPS -> compatible ( both are downloading )
HTTPS vs FTP -> compatible ( both are downloading )
DVD  vs NFS -> not compatible ( both are not downloading , but still not compatible )

zypp-logic/zypp-core/MirroredOrigin.h
```diff
@@ -316,7 +316,10 @@ namespace zypp {
     /*!
      * An authority and a mirror are not compatible if one is downloadable and the other is not
      */
-    bool isAuthorityCompatible(const OriginEndpoint& authority, const OriginEndpoint& mirror);
+    static bool isAuthorityCompatible(const OriginEndpoint& authority, const OriginEndpoint& mirror)
+    {
+      return (authority.schemeIsDownloading() == mirror.schemeIsDownloading());
+    }
```
bzeller, line 324: Ah you fixed the issue here, we could simply check if the schema is equal here as well, then not all places that call this function also need to check if the schemas are equal

bzeller: FYI this is optional, it does not block this PR from landing

danidoni: I think I fixed it.

PR #724, opened 2026-04-13
Can we haz cpp20

bzeller:
This adds support for C++20 for all our release targets

Depends on PR:  https://github.com/openSUSE/libzypp/pull/723


bzeller:
After extensive testing and ABI analysis using `libabigail` and `abidiff`, we have determined that enabling **C++20** for `libzypp` while maintaining a **C++11** consumer base (like YaST) is not feasible. This change triggers significant binary breaks that would destabilize the distribution.

### Technical Justification for Closure

#### 1. API Autoremoval (Reserved Keywords)
Several core symbols, including `zypp::sat::Solvable::conflicts()`, `provides()`, and `requires()`, utilize identifiers that are now **reserved keywords** in C++20. 
* These APIs were previously deprecated and are automatically removed or disabled by the headers when compiling with `-std=c++20` to avoid syntax errors. 
* Because these symbols disappear from the `.so` entirely, any binary compiled against older versions (like YaST) will fail to launch with `undefined symbol` errors.

#### 2. Vtable Corruption
`abidiff` detected unintended virtual function insertions and shifts in core classes such as `zypp::RepoInfo`. 
* Even without source changes, C++20’s stricter enforcement of `noexcept` specifications and changes to destructor emission rules alter the **virtual table (vtable) layout**. 
* YaST would consequently attempt to call functions at incorrect memory offsets, leading to immediate segmentation faults or "pure virtual function call" errors.

#### 3. Standard Library Layout Divergence (ODR Violations)
Analysis of `std::optional` and other containers shows internal structural shifts in `libstdc++` when toggled to C++20 mode. 
* Passing these objects between a C++11 caller and a C++20 library violates the **One Definition Rule (ODR)**. 
* The two binaries no longer agree on the internal state or memory footprint of these standard types, resulting in silent memory corruption.

#### 4. Calling Convention Shifts
The refined definition of "Aggregates" in C++20 alters how the compiler passes certain structs via CPU registers versus the stack. This creates a binary mismatch where the library and the application no longer agree on the machine-level protocol for communicating data.

### Conclusion

While the transition to C++20 offers significant internal advantages (such as Ranges and Concepts), the resulting "Binary Ghost" created by the standard mismatch is a critical risk for a foundational system library. 

To support C++20 in the future, we would need to:
1. Rename or refactor the deprecated symbols that conflict with C++20 keywords.
2. Ensure a synchronized rebuild of the entire SUSE management stack (YaST, Boost, etc.) using the same standard.

For the current release cycle, `libzypp` will remain on the existing stable standard to preserve ABI integrity.

PR #736, opened 2026-05-14
StringV: guard hasPrefix/hasPrefixCI against reading past the view end

### LINKED ISSUE #735: strv::hasPrefix/hasPrefixCI read past end of std::string_view when prefix is longer than view
## Summary

`zypp::strv::hasPrefix(std::string_view, std::string_view)` and its case-insensitive overload `hasPrefixCI` invoke `::strncmp` / `::strncasecmp` for `prefix_r.size()` bytes without first checking that `str_r.size() >= prefix_r.size()`. When the prefix is longer than the view, this reads past the end of the view's underlying buffer, which is undefined behavior.

Source (`zypp-logic/zypp-core/base/StringV.h`):

```cpp
inline bool hasPrefix( std::string_view str_r, std::string_view prefix_r )
{ return( ::strncmp( str_r.data(), prefix_r.data(), prefix_r.size() ) == 0 ); }

inline bool hasPrefixCI( std::string_view str_r, std::string_view prefix_r  )
{ return( ::strncasecmp( str_r.data(), prefix_r.data(), prefix_r.size()  ) == 0 ); }
```

`std::string_view` is not required to point at a NUL-terminated buffer, so `strncmp` cannot rely on a terminator to stop early. The companion `str::hasSuffix` in `zypp-logic/zypp-core/base/String.h` is implemented correctly because it short-circuits on `str_r.size() >= suffix_r.size()` first; the `strv::hasPrefix` overloads are missing this guard.

## Impact

* **Incorrect result.** When the view points into a longer underlying buffer and the prefix happens to match the bytes immediately after the view's end, `hasPrefix` returns `true` for a prefix that is strictly longer than the view. Concretely:

  ```cpp
  const char buf[] = "abcdef";
  std::string_view sv( buf, 3 );          // "abc"
  zypp::strv::hasPrefix( sv, "abcde" );   // returns true; should be false
  ```

* **Heap-buffer-overflow.** When the view sits at the end of an allocation, the extra bytes read by `strncmp` step outside the allocation. AddressSanitizer flags this as a heap-buffer-overflow:

  ```
  ==…==ERROR: AddressSanitizer: heap-buffer-overflow on address …
  READ of size 4 at … thread T0
      #0 strncmp …
      #1 hasPrefix(std::string_view, std::string_view) …
  ```

## Affected callers

`strv::hasPrefix` / `strv::hasPrefixCI` are called on untrimmed and trimmed views of curl response bytes in:

* `zypp-logic/zypp-curl/ng/network/curlmultiparthandler.cc` (`HTTP/`, `Content-Range:`, `Content-Type:`)
* `zypp-logic/zypp-curl/ng/network/request.cc` (`HTTP/`, `Location:`, `Content-Length:`)
* `zypp/zypp/Package.cc` (compares an `IdString::asStringView()` against another `asStringView()`)

For the curl callers the view is a substring of a libcurl-owned buffer after whitespace trimming; the size of that view is not guaranteed to be at least the size of the literal it is compared against. The first time the response payload happens to be shorter than e.g. `"Location:"`, the call invokes UB.

## Fix

Add the missing size guard, matching the style already used by `str::hasSuffix`:

```cpp
inline bool hasPrefix( std::string_view str_r, std::string_view prefix_r )
{ return( str_r.size() >= prefix_r.size() && ::strncmp( str_r.data(), prefix_r.data(), prefix_r.size() ) == 0 ); }

inline bool hasPrefixCI( std::string_view str_r, std::string_view prefix_r  )
{ return( str_r.size() >= prefix_r.size() && ::strncasecmp( str_r.data(), prefix_r.data(), prefix_r.size()  ) == 0 ); }
```

A PR with the fix and a regression test in `zypp/tests/zypp/StringV_test.cc` will be opened shortly.

officialasishkumar:
## Summary

`zypp::strv::hasPrefix(std::string_view, std::string_view)` and its case-insensitive overload pass `prefix_r.size()` straight to `::strncmp` / `::strncasecmp` without first checking that `str_r.size() >= prefix_r.size()`. Because `std::string_view` is not required to be NUL-terminated, when the prefix is longer than the view both calls read past the end of the view's underlying buffer (undefined behavior, flagged by AddressSanitizer as a heap-buffer-overflow). If the bytes following the view happen to match the rest of the prefix, the call also returns the wrong boolean result.

This PR adds the missing length short-circuit, matching the style already used by `str::hasSuffix` in `zypp-core/base/String.h`, and adds a regression test case in `StringV_test.cc`.

```cpp
inline bool hasPrefix( std::string_view str_r, std::string_view prefix_r )
{ return( str_r.size() >= prefix_r.size() && ::strncmp( str_r.data(), prefix_r.data(), prefix_r.size() ) == 0 ); }

inline bool hasPrefixCI( std::string_view str_r, std::string_view prefix_r  )
{ return( str_r.size() >= prefix_r.size() && ::strncasecmp( str_r.data(), prefix_r.data(), prefix_r.size()  ) == 0 ); }
```

Callers in the zyppng curl path (`curlmultiparthandler.cc`, `request.cc`) check curl-supplied header lines against literals like `"HTTP/"`, `"Location:"`, `"Content-Length:"`, `"Content-Range:"`, `"Content-Type:"` after trimming, so a trimmed view shorter than the literal would previously have triggered the over-read.

Closes #735

## Reproducer (AddressSanitizer)

```cpp
char *p = (char*)malloc(3);
p[0] = 'a'; p[1] = 'b'; p[2] = 'c';
std::string_view sv( p, 3 );
zypp::strv::hasPrefix( sv, "abcde" );  // heap-buffer-overflow, may also return true
```

```
==…==ERROR: AddressSanitizer: heap-buffer-overflow on address …
READ of size 4 at … thread T0
    #0 strncmp …
    #1 hasPrefix(std::string_view, std::string_view) …
```

After the fix the call returns `false` without entering `strncmp`.

## Test plan

- [x] Added `BOOST_AUTO_TEST_CASE(prefix)` in `zypp/tests/zypp/StringV_test.cc` covering:
  - empty prefix (always present),
  - plain prefix matches (including full string and case mismatch),
  - case-insensitive matches,
  - prefix strictly longer than the view (must return false; previously could be UB or wrong result),
  - a `std::string_view` that aliases a longer underlying buffer (the case that previously returned `true` for a too-long prefix because `strncmp` walked into the surrounding bytes).
- [x] Replicated the buggy and fixed implementations standalone under `g++ -std=c++17 -fsanitize=address` to verify ASan flags the original `strncmp` over-read and that the size-guarded version returns the correct result without reading past the view end.

### DIFF
--- a/zypp-logic/zypp-core/base/StringV.h
+++ b/zypp-logic/zypp-core/base/StringV.h
@@ -93,11 +93,11 @@ namespace zypp
 
     /** Return whether \a str_r has prefix \a prefix_r. */
     inline bool hasPrefix( std::string_view str_r, std::string_view prefix_r )
-    { return( ::strncmp( str_r.data(), prefix_r.data(), prefix_r.size() ) == 0 ); }
+    { return( str_r.size() >= prefix_r.size() && ::strncmp( str_r.data(), prefix_r.data(), prefix_r.size() ) == 0 ); }
 
     /** \overload Case insensitive */
     inline bool hasPrefixCI( std::string_view str_r, std::string_view prefix_r  )
-    { return( ::strncasecmp( str_r.data(), prefix_r.data(), prefix_r.size()  ) == 0 ); }
+    { return( str_r.size() >= prefix_r.size() && ::strncasecmp( str_r.data(), prefix_r.data(), prefix_r.size()  ) == 0 ); }
 
     ///////////////////////////////////////////////////////////////////
     namespace detail
--- a/zypp/tests/zypp/StringV_test.cc
+++ b/zypp/tests/zypp/StringV_test.cc
@@ -234,3 +234,49 @@ BOOST_AUTO_TEST_CASE(trimming)
   BOOST_CHECK_EQUAL(  trim(" \t1\t ", " "),  "\t1\t" );
   BOOST_CHECK_EQUAL(  trim(" \t1\t ", ""),  " \t1\t " );
 }
+
+BOOST_AUTO_TEST_CASE(prefix)
+{
+  // The empty prefix is always present.
+  BOOST_CHECK( hasPrefix( "", "" ) );
+  BOOST_CHECK( hasPrefix( "abc", "" ) );
+  BOOST_CHECK( hasPrefixCI( "", "" ) );
+  BOOST_CHECK( hasPrefixCI( "abc", "" ) );
+
+  // Plain prefix matching.
+  BOOST_CHECK( hasPrefix( "abc", "a" ) );
+  BOOST_CHECK( hasPrefix( "abc", "ab" ) );
+  BOOST_CHECK( hasPrefix( "abc", "abc" ) );
+  BOOST_CHECK( ! hasPrefix( "abc", "b" ) );
+  BOOST_CHECK( ! hasPrefix( "abc", "Ab" ) );
+
+  // Case insensitive matching.
+  BOOST_CHECK( hasPrefixCI( "abc", "A" ) );
+  BOOST_CHECK( hasPrefixCI( "abc", "AB" ) );
+  BOOST_CHECK( hasPrefixCI( "abc", "ABC" ) );
+  BOOST_CHECK( hasPrefixCI( "ABC", "abc" ) );
+  BOOST_CHECK( ! hasPrefixCI( "abc", "B" ) );
+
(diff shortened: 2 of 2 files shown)


mlandres:
Thanks!

PR #743, opened 2026-06-16
Fix duplicated http header

mlandres:
[bsc#1268321](https://bugzilla.suse.com/show_bug.cgi?id=1268321)


zypp-logic/zypp-curl/ng/network/networkrequestdispatcher.h
```diff
@@ -9,9 +9,12 @@
 #include <unordered_map>
 
 #include <zypp-curl/ng/network/networkrequesterror.h>
+#include <zypp/media/MediaNetworkCommonHandler.h>
```
bzeller: This pulls legacy code into NG namespace. Thats a no go, after the seperation this will break

zypp-logic/zypp-curl/ng/network/networkrequestdispatcher.cc
```diff
@@ -464,17 +464,19 @@ const std::string &NetworkRequestDispatcher::agentString() const
...
-    if ( auto i = d->_customHeaders.find( host ); i != d->_customHeaders.end() ) {
-      if ( auto v = i->second.find( headerName ); v != i->second.end() )  {
-        i->second.erase (v);
-      }
-      if ( i->second.empty() )
-        d->_customHeaders.erase(i);
+  d->_customHeaders[host][headerName] = value;
+}
+
+void NetworkRequestDispatcher::unsetHostSpecificHeader( const std::string &host, const std::string &headerName )
```
bzeller, line 470: Why a unset function instead of using set() with a empty value?

bzeller (changes requested):

bzeller (approved):

zypp-logic/zypp-curl/ng/network/networkrequestdispatcher.h
mlandres: fixed

PR #747, opened 2026-07-07
zypp-curl: Enable HTTP3 if supported by curl and wait for confirmation on multiplexing support

crrodriguez:
After this patch one can confirm connections to the fastly CDN use HTTP3 and multiplexing works over a single UDP "connection" instead of opening a multitude of them and later catching up. (multiplexing is enabled by default in libcurl)

Other mirrors that support HTTP2 and "confirm" multiplex also do over TCP.

Most other mirrors work over http 1.1 where there is no multiplexing and curl no longer support pipelining so multiple connections are attempted.

side effect: Tiny delay to start download.

### DIFF
--- a/zypp-logic/zypp-curl/ng/network/request.cc
+++ b/zypp-logic/zypp-curl/ng/network/request.cc
@@ -120,7 +120,9 @@ namespace zyppng {
       setCurlOption( CURLOPT_NOPROGRESS, 0L);
       setCurlOption( CURLOPT_FAILONERROR, 1L);
       setCurlOption( CURLOPT_NOSIGNAL, 1L);
+#if CURLVERSION_AT_LEAST(8,16,0)
       setCurlOption( CURLOPT_PIPEWAIT, 1L);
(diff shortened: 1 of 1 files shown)


mlandres:
Looks like it works on TW only. Code 15 testsuites fail.

crrodriguez:
> Looks like it works on TW only. Code 15 testsuites fail.

There seems to be an issue with older curl versions, setting the http3 bit is probably not the cause..since it is either not present or a no-op on old systems that do not have the required openssl/ngtcp/nghttp versions.however it is probably the pipewait thing.
I-ll check whats going on.

crrodriguez:
my bad.. actually. I assumed setCurlOption didnt care about the return value.. of course setopt is gonna fail in a curl release that has the HTTP3 bit but no support for it. it will return protocol not supported and exception will be thrown.. that happens when you dont read the function defintion and only the declaration..

crrodriguez:
OK, all green now.

PR #750, opened 2026-07-18
fix character width calculation

YoukouTenhouin:
In certain languages (e.g. CJK languages), some characters may have different width than the number of UTF8 code points (e.g. CJK characters usually take double the width). Failure to taken that into account will result in incorrect string width calculation. This might cause the progress bar to overflow line width, results in spamming the terminal.

Use `mbs_width` to calculate character width fixes the issue.


mlandres:
@YoukouTenhouin zypp-tui is a library. You actually can not tell whether a public (installed) header is used or unused without checking all applications using the lib. The Utf8.h must stay.

YoukouTenhouin:
@mlandres Dropped the clean up commit. Thanks for clarifying this!

mlandres:
Thanks @YoukouTenhouin.

`mbs_width` and `mbs_substr_by_width` are built using `MbsIterator` which apparently handles the CJK chars correctly. 

`utf8::string` itself should probably be based on using `MbsIteratorNoSGR`. The class promises to transparently handle embedded ANSI color sequences as well. Your fix replaces `utf8::string` by `MbsIterator`  based functions. So embedded ANSI color sequences would probably no longer be counted correctly (as 0-length).

IMO the correct fix would be to re-implement `utf8::string::size()`  and `substr()` similar to `mbs_width` and `mbs_substr_by_width` but using an `MbsIteratorNoSGR`. For `size` this is fairly straight forward. For `substr` one must keep in mind that the iterator skips any leading SGR upon construction. A substr starting at 0 must not forget to include them.

PR #752, opened 2026-07-28
More changes to support the MCP server for zypp

bzeller:


zypp/zypp/misc/LoadTestcase.cc
```diff
@@ -211,12 +217,26 @@ namespace zypp::misc::testcase {
...
+
+    // ── Extract lock/keep nodes from all trials into the setup ─────────────────
+    // Done as a post-parse pass so neither parser needs to be modified.
+    // Locks are pool constraints that belong to the environment, not the job.
+    for ( const auto & trial : _pimpl->_trials )
+      for ( const auto & node : trial.nodes() )
+        if ( node.name() == "lock" || node.name() == "keep" )
+          _pimpl->_setup.data().locks.push_back( { node.name(), node.properties() } );
+
+    return true;
```
bzeller: I'm not sure this is correct. We usually have only one trial anyway but combining the locks would be wrong if multiple trials are there

mlandres (approved):

PR #757, opened 2026-08-10
Do not join the log thread in a forked child

bmwiedemann:
Note: ClaudeAI came up with this and it looks reasonable to me. Give it extra careful review+testing.
Check it live at https://build.opensuse.org/project/show/home:bmwiedemann:zypperd

===

A forked child inherits the `LogThread` singleton object, but not the thread itself. When such a child exits via `exit()` instead of `_exit()`, the LogThread destructor runs from `__run_exit_handlers` and `join()` blocks forever on a thread that does not exist in this process.

Observed with a fork-per-request zypper daemon (zypperd): a worker child that received `SIGPIPE` went through `Zypper::immediateExit -> exit()` and hung in the futex wait inside ~LogThread.

Remember the pid that created the thread and detach instead of joining when the destructor runs in a different process.

### DIFF
--- a/zypp-logic/zypp-core/base/LogControl.cc
+++ b/zypp-logic/zypp-core/base/LogControl.cc
@@ -110,6 +110,15 @@ namespace zypp
 
     void stop () {
       _stopSignal.notify();
+      if ( _ownerPid != getpid() )
+      {
+        // We are in a forked child: the worker thread only exists in the
+        // parent, join() would block forever (e.g. a fork server calling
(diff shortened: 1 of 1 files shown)


bzeller:
We probably did not see this happen because all our internal workers fork+exec

mlandres:
In my tests I don't see this blocking if the child calls `exit()` but it actually does not matter much. 
Zypp is not designed for this kind of use and my 1st child exiting closes the parent's log. I don't know hat else could be messed up. From this POV it might be good that it blocks.

